PR #16, opened 2023-01-16
Align 35 prisoner profile overview

jamesgrant-moj:


server/views/pages/overviewPage.njk
```diff
@@ -0,0 +1,16 @@
+<div class="govuk-grid-row">
+    <div class="govuk-grid-column-three-quarters">
+        <h1>Overview</h1>
+    </div>
+</div>
+
+<div class="govuk-grid-row">
+    <div class="govuk-grid-column-three-quarters">
+        {% include "../partials/mainBody.njk" %}
```
davidmiddletonmoj, line 9: You've moved the partials into an `overviewPage` subfolder, but the references are still to the "old" partials.

Also as per Dan's Slack comment, the "old" partials are still there, so we've got duplicates which can cause issues like this.

davidmiddletonmoj (approved):
👍

davidmiddletonmoj (approved):
👍

PR #20, opened 2023-01-23
ALIGN-39 - Mini summary macro - box style 1 and 2

jamesgrant-moj:
This PR includes the mini summary macro for box style 1 and 2 (ALIGN-43)

PR #21, opened 2023-01-24
ALIGN-41 - Summary card component

danielburnley:
A macro for the summary card component along with the sections on the home page which use it

This has some duplication with #20 - so there'll be a follow-up PR to address this and combine the CSS into one that can be used across both


jamesgrant-moj:
Looks good Dan. The CSS needs a small tweak to improve responsiveness. The summary cards for Today's schedule and non associations needs adjustment for screen sizes around 670 pixels. I tested on chrome Version 109.0.5414.87 using a mac although I don't think it is browser related. But other than that it all LGTM.

assets/sass/components/_summary-card.scss
```diff
@@ -0,0 +1,20 @@
+.hmpps-summary-card {
+  border: 1px solid black;
```
davidmiddletonmoj, line 2: There doesn't seem to be a callout for border colour on the Figma design, but inspecting it I think it's #505A5F rather than "black"

Would be good to use the `govuk-colour` function to define these - #505A5F would be `dark-grey`

And finally, just my OCD... better to remove commented out lines if they are not needed any more.

jamesgrant-moj: Good spot. Please replace black with  govuk-colour("dark-grey") or whatever colour is needed.
https://design-system.service.gov.uk/styles/colour/

assets/sass/components/_summary-card.scss
```diff
@@ -0,0 +1,20 @@
...
+  // margin-bottom: 10px;
+
+  .hmpps-summary-card__header {
+    // padding: 0 15px;
+    background-color: #F3F2F1;
```
jamesgrant-moj, line 7: Please add this govuk colour too. Or it can be done in the CSS clean up task.
govuk-colour("light-grey")

jamesgrant-moj (approved):
LGTM @danielburnley

PR #22, opened 2023-01-24
ALIGN-41 - Align card header/bodies with summary macro

danielburnley:
A follow up to both #21 and #20 combining the two ways of doing the card components into one


jamesgrant-moj (approved):
LGTM

PR #25, opened 2023-01-27
ALIGN-44: Overview page tidy up

davidmiddletonmoj:
WIP


danielburnley (commented):
I realise this is a WIP but this looks good so far 🎉 it doesn't conflict with #22 but it might be worth keeping both this branch and that multiple short-lived branches to avoid conflicts if we have crossover whilst clean up is happening

jamesgrant-moj (approved):
LGTM!

PR #26, opened 2023-01-27
ALIGN-45 change location condition

jamesgrant-moj:


server/data/hmppsAuthClient.ts
```diff
@@ -47,6 +48,17 @@ export default class HmppsAuthClient {
     return new RestClient('HMPPS Auth Client', config.apis.hmppsAuth, token)
   }
 
+  getUserLocations(token: string): Promise<Location[]> {
+    return HmppsAuthClient.restClient(token)
+      .get({ path: '/api/users/me/locations' })
```
davidmiddletonmoj, line 53: I think this is the wrong endpoint... seems like it returns locations _within_ the prison, whereas we want the prisons themselves, which are found at `/api/users/me/caseLoads`, returning an array of caseload (prison) models,
```
 {
    "caseLoadId": "MDI",
    "description": "Moorland Closed (HMP & YOI)",
    "type": "INST",
    "caseloadFunction": "GENERAL",
    "currentlyActive": false
  }
```

server/data/hmppsAuthClient.ts
```diff
@@ -47,6 +48,17 @@ export default class HmppsAuthClient {
     return new RestClient('HMPPS Auth Client', config.apis.hmppsAuth, token)
   }
 
+  getUserLocations(token: string): Promise<Location[]> {
```
davidmiddletonmoj, line 51: Probably should use the same terminology as the API, so `getUserCaseLoads`

server/data/localMockData/locations.ts
```diff
@@ -0,0 +1,58 @@
+type Location = {
```
davidmiddletonmoj, line 1: I think this is the wrong model - should be `CaseLoad` which represents a prison.
e.g.
```
  {
    "caseLoadId": "MDI",
    "description": "Moorland Closed (HMP & YOI)",
    "type": "INST",
    "caseloadFunction": "GENERAL",
    "currentlyActive": false
  }
```

server/middleware/populateCurrentUser.ts
```diff
@@ -20,3 +20,41 @@ export default function populateCurrentUser(userService: UserService): RequestHa
     }
   }
 }
+
+export function getUserLocations(userService: UserService): RequestHandler {
```
davidmiddletonmoj, line 24: Naming should be consistent with CaseLoad being used in place of Location

Comment applies across multiple files/lines.

jamesgrant-moj:
@davidmiddletonmoj thanks for taking a look.

I have added the caseloads api endpoint. 

I have left the locations work in there as we will still need it for when we rebuild the locations details page.
https://digital-dev.prison.service.justice.gov.uk/prisoner/G6123VU/cell-history

server/data/localMockData/caseLoad.ts
```diff
@@ -0,0 +1,34 @@
+export type CaseLoad = {
```
danielburnley, line 1: Do we want these types to live in `server/data/localMockData`? If we're building these to subsequently return the same shape from the real data I would expect to find these somewhere else otherwise these look like they'd only be the shape for mock data. We could put them potentially alongside the actual auth client or elsewhere and leave these files just for the data

danielburnley (commented):
Just a smol comment about the location of the types we're passing around

server/data/localMockData/caseLoad.ts
davidmiddletonmoj: Good point. Maybe use `server/interfaces`?

jamesgrant-moj: Good idea.

danielburnley (approved):
LGTM so far - I'm not a greattt fan of PRs being open for ages for minor things, but stuff we discussed yesterday that'd be worth changing as we continue to work on this page:

- Would be good to move the `.then`s to `async/await` the next time you might be in that area of code, just for consistency (& IMO readability)
- Moving the calling of the services to the routes rather than middleware and passing the objects in there

Otherwise this looks good to me and feel free to merge it if you're finished 🎉

PR #33, opened 2023-02-01
Setting up prison API access

brightonsbox:


server/authentication/auth.ts
```diff
@@ -40,9 +40,8 @@ function init(): void {
       customHeaders: { Authorization: generateOauthClientToken() },
     },
     (token, refreshToken, params, profile, done) => {
-      console.log(token)
```
brightonsbox, line 43: Need to make sure this is removed in your current branch - don't want to be logging out user tokens anywhere

jamesgrant-moj: Yes will do Jon.

PR #42, opened 2023-02-08
CDPS-47-banner-profile-photo

jamesgrant-moj:


server/mappers/headerMappers.ts
```diff
@@ -3,6 +3,9 @@ import { tabLinks } from '../data/profileBanner/profileBanner'
 import { AlertFlagLabel } from '../interfaces/alertFlagLabels'
 import { alertFlagLabels } from '../data/alertFlags/alertFlags'
 
+export const placeHolderImagePath = '/assets/images/prisoner-profile-photo.png'
+export const subStr = 'localhost:3000'
```
danielburnley, line 7: This could do with being a more descriptive name  rather than `subStr` - given it's used for logic in the `pageService` and only there I think it'd be better if it was moved out of this file altogether

server/services/pageService.ts
```diff
@@ -3,16 +3,34 @@ import { Prisoner } from '../interfaces/prisoner'
...
-      _res.render(template, {
-        ...mapHeaderData(prisonerData),
-        ...pageData,
-      })
+  async renderPage<T>(_res: Response, prisonerNumber: string, template: string, pageData: T) {
+    const prisonerSearch = new PrisonerSearchClient(new TokenStore(createRedisClient({ legacyMode: false })))
+    const prisonerData: Prisoner = await prisonerSearch.getPrisonerDetails(_res.locals.user.token, prisonerNumber)
+
+    let imagePath: string
+    if (_res.req.rawHeaders[1].includes(subStr)) {
```
danielburnley, line 15: following on from above - i'd probably actually move this entire bit to a private method in the class called `isLocalhost` or something that describes what it's checking here and do `if(isLocalhost(_res))` so it's clear what's happening here, with `isLocalhost` being something that does this check, probably with `subStr` hardcoded as just `localhost:3000` if it doesnt need to be used elsewhere

jamesgrant-moj: Thanks for taking a look @danielburnley. I have updated the PR with your change suggestions.

So currently the placeholder grey image displays when running the app locally.
/assets/images/prisoner-profile-photo.png

To prevent running the API call which will always return 404 for the moment(when running locally). The logic for the profile photo is not setup to work locally from dev. This is why I am checking for localhost.

If the localhost check fails the image api will catch block will handle the failure and render the placeholder image path.

My assumption is the real image will display on dev, but this is still an unknown as this is the first image implementation on this repo rendering an image from calling APIs. I plan to test on dev and see how this logic behaves.

danielburnley: Yeah that makes sense - I wouldn't worry about making it too generic, it might just be in the future we just accept the 404's but for testing/showing off the new functionality this makes sense 👍🏻

danielburnley (approved):

PR #44, opened 2023-02-09
CDPS-47 banner profile photo

jamesgrant-moj:


server/data/prisonApiClient.ts
```diff
@@ -30,12 +30,12 @@ export default class PrisonApiClient {
     }) as Promise<CaseLoad[]>
   }
 
-  async getPrisonerImage(offenderNumber: string, fullSizeImage: boolean): Promise<string> {
+  async getPrisonerImage<T>(offenderNumber: string, fullSizeImage: boolean): Promise<T> {
```
danielburnley, line 33: Does this/the api client need to be a generic? Or can it just be `Readable`? I'm not sure if we'll be calling it expecting to cast it to a different type, so correct me if im wrong :D but if so, you can remove this `<T>` from here and the other places and just keep `return as Readable`

jamesgrant-moj:
@danielburnley thanks for having a look. I updated the type.

danielburnley (approved):

PR #48, opened 2023-02-13
CDPS-58: mini summary boxes

davidmiddletonmoj:


server/services/pageService.ts
```diff
@@ -1,16 +0,0 @@
-import { Response } from 'express'
```
jamesgrant-moj, line 1: @davidmiddletonmoj @danielburnley @brightonsbox, are we all happy to move away from the page service? In hindsight is there any value in it?

jamesgrant-moj (commented):
LGTM - waiting for e2e tests

jamesgrant-moj (approved):

server/services/pageService.ts
danielburnley: I think it makes sense for now - If we were calling the prisoner search service in the main route for the page itself we'd be doing a redundant call here. I could see moving it to the routes to end up needing a refactoring as we always call the header data, but in my opinion we should just let evolve naturally as our service does rather than try figure it out now given we're not sure what will/won't be needed by other routes yet

server/routes/index.ts
```diff
@@ -21,11 +23,17 @@ export default function routes(service: Services): Router {
...
-    const pageService = new PageService()
-    const prisonApi = new PrisonApiRestClient(res.locals.user.token)
+    const prisonerSearchClient = new PrisonerSearchClient(res.locals.clientToken)
+    const prisonerData: Prisoner = await prisonerSearchClient.getPrisonerDetails(req.params.prisonerNumber)
+
+    const prisonApi = new PrisonApiRestClient(res.locals.clientToken)
     const overviewPageService = new OverviewPageService(prisonApi)
-    const overviewPage = await overviewPageService.get(req.params.prisonerNumber)
-    pageService.renderPage<OverviewPage>(res, req.params.prisonerNumber, 'pages/index', overviewPage)
+    const overviewPageData = await overviewPageService.get(prisonerData)
```
jamesgrant-moj, line 31: @davidmiddletonmoj, is this suppose to be prisoner Data or req.params.prisonerNumber?

davidmiddletonmoj: It's `prisonerData` now - the method pulls out a few bits of info from this object

jamesgrant-moj (approved):
LGTM

PR #49, opened 2023-02-14
CDPS-39-photo-page

jamesgrant-moj:


server/routes/index.ts
```diff
@@ -30,9 +33,33 @@ export default function routes(service: Services): Router {
...
+    })
+  })
+
+  get('/prisoner/:prisonerNumber/image', async (req, res, next) => {
+    const prisonerSearchClient = new PrisonerSearchClient(res.locals.clientToken)
+    const prisonerData: Prisoner = await prisonerSearchClient.getPrisonerDetails(req.params.prisonerNumber)
+
+    const prisonApi = new PrisonApiRestClient(res.locals.clientToken)
+    const overviewPageService = new OverviewPageService(prisonApi)
+    const overviewPageData = await overviewPageService.get(prisonerData)
```
davidmiddletonmoj, line 53: I don't think we need to do all the API calls for the overview page as this just displays the image.
From the template, seems like you just need the prisonerNumber (which is in the request param), photoType and the name, which could come from the first call to `getPrisonerDetails`?

(Maybe something to think about for the future would be if we want to store any session state, like prisoner name, or even just be able to pass objects from one page to the next?)

davidmiddletonmoj (commented):
Maybe one for Dom/Eddie - I think we need a hover state for the photo link. Not sure if other services have examples of this?

davidmiddletonmoj:
I think the back link should be active and be "Back to the profile page"

jamesgrant-moj:
@davidmiddletonmoj Thanks for the feedback. Ready for another review.

davidmiddletonmoj (approved):

PR #51, opened 2023-02-15
CDPS-60 - personal details

jamesgrant-moj:


server/services/overviewPageService.ts
```diff
@@ -32,6 +34,75 @@ export default class OverviewPageService {
     }
   }
 
+  private getPersonalDetails(prisonerData: Prisoner): PersonalDetails {
+    const personalDetailsMain = [
```
davidmiddletonmoj, line 38: Could extract this into a mock object so it can also be used for unit and e2e testing.

server/services/overviewPageService.ts
```diff
@@ -20,6 +21,7 @@ export default class OverviewPageService {
     const nonAssociations = await this.getNonAssociations(prisonerNumber)
     const miniSummaryGroupA = await this.getMiniSummaryGroupA(prisonerNumber, bookingId)
     const miniSummaryGroupB = await this.getMiniSummaryGroupB(currentIncentive, bookingId)
+    const personalDetails = this.getPersonalDetails(prisonerData)
```
danielburnley, line 24: We could do with a test around this to make sure it's properly creating the return value correctly as it's doing some data manipulation/mapping

davidmiddletonmoj (approved):

danielburnley:
In addition to ^ - might be good to get this in the cypress tests as well

PR #59, opened 2023-02-21
CDPS-59 staff contacts component

jamesgrant-moj:


server/services/overviewPageService.test.ts
```diff
@@ -27,24 +33,42 @@ describe('OverviewPageService', () => {
...
+
   beforeEach(() => {
     prisonApiClient.getNonAssociationDetails = jest.fn(async () => nonAssociationDetailsDummyData)
     prisonApiClient.getEventsScheduledForToday = jest.fn(async () => dummyScheduledEvents)
   })
 
   describe('Non-associations', () => {
     it.each(['ABC123', 'DEF321'])('Gets the non-associations for the prisoner', async (prisonerNumber: string) => {
-      const overviewPageService = new OverviewPageService(prisonApiClient)
+      const overviewPageService = new OverviewPageService(
```
danielburnley, line 57: I wonder if we could just move the construction of the overview page service to at least a function - so line 62 would become:
```await service().get({prisonerNumber} as Prisoner)```
where `service()` is a function that returns
```
new OverviewPageService(prisonApiClient, ...)
```

Might clean up these tests a bit. Not sure if it'll work though

File: server/data/allocationManagerApiClient.ts [hunk 2]
```diff
@@ -0,0 +1,19 @@
+import RestClient from './restClient'
+import config from '../config'
+import { Pom } from '../interfaces/pom'
+
+export default class AllocationManagerClient {
```
danielburnley, line 5: This should be named `AllocationManagerApiClient` to match the filename, and also needs to implement `AllocationManagerClient`

server/services/overviewPageService.ts
```diff
@@ -46,6 +66,67 @@ export default class OverviewPageService {
     }
   }
 
+  public async getStaffContacts(prisonerData: Prisoner): Promise<StaffContacts> {
+    const { bookingId } = prisonerData
+    const [offenderContacts, allocationManager, offenderKeyWorker, keyWorkerSessions] = await Promise.all([
+      this.prisonApiClient.getOffenderContacts(prisonerData.bookingId),
+      this.allocationManagerClient.getPomByOffenderNo(prisonerData.prisonerNumber).catch(() => undefined),
```
danielburnley, line 73: Is there a reason this one has the `.catch()` where the others do not? it causes that bit below (lines 88-92) where you have to contend with it being `undefined` and also of `any` type so I wondered if it was necessary to catch it here where we don't any of the others.

server/utils/utils.ts
```diff
@@ -94,3 +94,24 @@ export const formatMoney = (val: number, emptyState: string = undefined, currenc
...
+    .join('&')
+
+export const getNamesFromString = (string: string): string[] =>
+  string &&
+  string
+    .split(', ')
+    .reverse()
+    .join(' ')
+    .split(' ')
+    .map(name => properCaseName(name))
```
danielburnley, line 117: This seems like it specifically needs a particular input string rather than just "get names from a string" so i wonder if it belongs in utils or if it could be just a local function to the overview page method.

Whether we keep it here or not, it'd probably be good to get some tests that test this particular bit of logic though as it seems important

danielburnley, line 117: Similar to my other comment - it'd be good to make sure we have tests around these functions as we'd want to make sure we don't accidentally break any of this if they were to be edited in the future

server/services/overviewPageService.test.ts
```diff
@@ -130,20 +178,47 @@ describe('OverviewPageService', () => {
...
+
+      const overviewPageService = new OverviewPageService(
+        prisonApiClient,
+        allocationManagerApiClient,
+        keyWorkerApiClient,
+      )
+      const res = await overviewPageService.get({ ...PrisonerMockDataB, prisonerNumber, bookingId } as Prisoner)
+      expect(res.staffContacts).toEqual(StaffContactsMock)
+    })
+  })
```
danielburnley, line 204: This test is good for the "everything is Ok" route, but I think there's some logic within the service which handles a lot of edge cases/missing data cases that'd be good to make sure it works so that we don't have to manually check it every time (e.g. Logic around no case notes)

jamesgrant-moj: Thanks for taking a look @danielburnley. I'll write some more tests. The utils are from the digital prison services repo, it would be good to wrap them in tests.

danielburnley (approved):

PR #62, opened 2023-02-24
CDPS-57: overview statuses data

davidmiddletonmoj:


server/services/overviewPageService.test.ts
```diff
@@ -214,4 +231,113 @@ describe('OverviewPageService', () => {
       expect(evening[1].endTime).toEqual('19:00')
     })
   })
+
+  describe('getStatuses', () => {
```
danielburnley, line 235: Nothing to be done for now - but i do wonder if this ends up being it's own service if we get much more statuses because that's going to be a lot 😅

server/services/overviewPageService.ts
```diff
@@ -266,6 +269,11 @@ export default class OverviewPageService {
   private async getMiniSummaryGroupB(currentIncentive: Incentive, bookingId: number): Promise<MiniSummary[]> {
     const assessments = await this.prisonApiClient.getAssessments(bookingId)
 
+    if (!Array.isArray(assessments)) {
+      // TODO handle api call returning error???
```
danielburnley, line 273: 👀 Is this intentionally left here for now?

danielburnley (approved):
LGTM - not sure if that TODO is left there intentionally, if so go ahead and merge 🎉

PR #72, opened 2023-03-13
CDPS-44 - Personal page personal details card

danielburnley:


server/services/personalPageService.ts
```diff
@@ -10,36 +12,64 @@ export default class PersonalPageService {
   }
 
   public async get(prisonerData: Prisoner): Promise<PersonalPage> {
-    // const inmateDetail = await this.prisonApiClient.getInmateDetail(prisonerData.bookingId)
-    const toFullName = (firstName: string, middleNames: string, lastName: string) =>
-      [firstName, middleNames, lastName].filter(s => s !== undefined).join(' ')
+    const inmateDetail = await this.prisonApiClient.getInmateDetail(prisonerData.bookingId)
```
davidmiddletonmoj, line 15: Could do these in parallel with `Promise.all` to improve performance

server/views/partials/personalPage/languages.njk
```diff
@@ -0,0 +1,42 @@
+{% from "../../macros/summaryListRow.njk" import summaryListRow %}
+{% macro otherLanguagesText(language) %}
+  {% if language.canSpeak and language.canWrite %}
```
davidmiddletonmoj, line 3: Do we need to have a `and not <the_other_canSomething>` for the first 3 conditions? i.e. what if all 3 are `true`?

davidmiddletonmoj (approved):
Some minor comments, but approved to keep things moving

server/views/partials/personalPage/languages.njk
danielburnley: Ah yeah you're right - I'll add that in in the next PR

PR #76, opened 2023-03-17
CDPS:82 - Alerts pagination

davidmiddletonmoj:


integration_tests/pages/alertsPage.ts
```diff
@@ -8,4 +8,28 @@ export default class AlertsPage extends Page {
...
+
+  paginationHeaderPageLink = (): PageElement =>
+    cy.get(
+      '.hmpps-paged-list-header .moj-pagination__item:not(.moj-pagination__item--active):not(.moj-pagination__item--previous):not(.moj-pagination__item--next)',
+    )
+
+  paginationFooterPageLink = (): PageElement =>
+    cy.get(
+      '.hmpps-paged-list-footer .moj-pagination__item:not(.moj-pagination__item--active):not(.moj-pagination__item--previous):not(.moj-pagination__item--next)',
+    )
```
danielburnley, line 34: I wonder if this could be cleaned up to be its own object so it can then be used across multiple - so rather than separate methods its something like:

```typescript
pagination = () => {
  header: () => {}
  summaryHeader: () => {}
  // etc 
}
```

then we can move the object to another file so it ends up being `pagination = () => paginationObject`

danielburnley (approved):
LGTM - small suggestion on some clean up but given you're moving forward with this after you can always do it in a subsequent PR

integration_tests/pages/alertsPage.ts
dmpro: Good idea... will implement with next PR

PR #83, opened 2023-03-23
CDPS-44 - Display address data from the API on the personal page

danielburnley:


server/interfaces/pages/personalPage.ts
```diff
@@ -49,8 +51,17 @@ export interface PropertyItem {
   location: string
 }
 
+export interface Addresses {
+  address: Address
+  comment: string
+  addressTypes: string[]
```
jamesgrant-moj, line 57: @danielburnley do you think it is worth sticking this in an enum?

danielburnley: Hmmm - I'm not sure if they'd be useful for this as we don't query for the address types anywhere, but if we did in the future it would definitely be worth it. I'll see if it makes sense as I do this next address related bit of work 🎉

PR #86, opened 2023-04-03
CDPS-83: Alerts filter

davidmiddletonmoj:


server/controllers/alertsController.ts
```diff
@@ -0,0 +1,44 @@
...
+ */
+export default class AlertsController {
+  private prisonerSearchService: PrisonerSearchService
+
+  private alertsPageService: AlertsPageService
+
+  constructor(clientToken: string) {
+    this.prisonerSearchService = new PrisonerSearchService(clientToken)
+    this.alertsPageService = new AlertsPageService(clientToken)
+  }
```
danielburnley, line 18: IMO if we're instantiating controllers it'd be good to do dependency inversion at the same time and pass in the services it's dependent on. Otherwise tests that want to control the responses of these can get a bit more messy and I've found jests module mocking a bit hit and miss. 

This also means we need to make interfaces for the services so that the class is dependent on those rather than the actual implementations, otherwise we can't pass them in.

server/services/alertsPageService.test.ts
```diff
@@ -2,21 +2,19 @@ import { Prisoner } from '../interfaces/prisoner'
...
   let getInmateDetailsSpy: jest.SpyInstance
   let getAlertsSpy: jest.SpyInstance
   let prisonerData: Prisoner
   let alertsPageService: AlertsPageService
 
   beforeEach(() => {
-    prisonApiClient = new PrisonApiRestClient(null)
     prisonerData = { bookingId: 123456, firstName: 'JOHN', lastName: 'SMITH' } as Prisoner
-    alertsPageService = new AlertsPageService(prisonApiClient)
+    alertsPageService = new AlertsPageService(null)
```
danielburnley, line 17: We could do with making a generic prison api mock (we have one made manually in 3 of the other services) - this way you can avoid this and make some of the setup a bit cleaner, e.g.: 

where you have the `getInmateDetailsSpy` below you can change it to:
`prisonApiClient.getInmateDetail = jest.fn(async () => ({return: 'value'})`

danielburnley (commented):
Introducing a controller layer seems like a fine idea to me - as we get more and more pages (I'm thinking the stuff like "Add a case note" and some of the other sub pages) we're going to be doing a bit more "grab stuff from the request object and put it into something" that wouldn't make sense to do in the _service_ but would bloat the routes if we keep adding to them.

I've made a couple comments around the testing, as its a change to how we've been doing the existing services - but it's pretty minor in terms of actual code change. I think it'd be worthwhile inverting dependencies and the like as it makes things a lot easier to parse in the tests IMO as you can set things up a bit nicer without needing to do the more verbose jest.mockimplementation etc etc. We do need to do a bit of clean up for the prison api client mocks that we keep adding to, but I'm happy to pick that up as part of my work if you just copy pasted the existing for now cos I need to add to it anyway 😅

package-lock.json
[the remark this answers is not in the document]
davidmiddletonmoj: Yep, exactly that. I removed my node_modules and package-lock and rebuilt.

jamesgrant-moj (commented):
LGTM @davidmiddletonmoj. I like all the changes. I agree with @danielburnley comments. This will come in handy especially when we start building the sub pages.

.eslintrc.json
```diff
@@ -26,13 +26,49 @@
...
+      "parser": "@typescript-eslint/parser",
+      "files": ["**/*.test.ts"],
       "excludedFiles": "*.js",
       "extends": [
         "plugin:@typescript-eslint/eslint-recommended",
         "plugin:@typescript-eslint/recommended",
         "plugin:prettier/recommended"
       ],
       "rules": {
+        "dot-notation": "off",
```
jamesgrant-moj, line 70: I am happy with this, was there a use case that convinced you to turn this off? We can always challenge the use of "any types" during the PR process to minimise it.

davidmiddletonmoj: This change is just for `*.test.ts` files as those rules specifically were preventing me from spying on private methods.
For application code, I think it's good to avoid using `any` still and those rules are still in place.

danielburnley (approved):

PR #90, opened 2023-04-14
CDPS-113: Setting up production environment

brightonsbox:


.env
davidmiddletonmoj, line 1: Not sure how this sneaked in here? Should be gitignored ?

davidmiddletonmoj (approved):
LGTM

PR #93, opened 2023-04-17
CDPS-85 offences

jamesgrant-moj:


server/utils/utils.ts
```diff
@@ -238,6 +238,9 @@ export const generateListMetadata = (
   }
 }
 
+export const formatCurrency = (number: number, currency: string): string =>
```
davidmiddletonmoj, line 241: There is already a `formatMoney` helper - is this needed in addition to that?

server/views/pages/offences.njk
```diff
@@ -0,0 +1,42 @@
...
+          iconFallbackText: "Warning"
+        }) }}
+      {% endif %}
+    </div>
+  </div>
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-one-quarter">
+      {% if courtCase.sentenceTerms.length == 0 %}
+        <div class="hmpps-sidebar-single-link">
+          <h4 class="govuk-!-margin-top-0">Court cases and offences</h4>
```
davidmiddletonmoj, line 19: Shouldn't skip heading levels - i.e. going from `h1` to `h4` here. Should go in sequence through the page source order, h1, h2, h3, etc
Can use govuk heading styles to change the size if needed, e.g. `govuk-heading-m` or `govuk-heading-s`

server/views/partials/offencesPage/courtCaseContent.njk
```diff
@@ -0,0 +1,98 @@
...
+    <ul class="govuk-list govuk-list--bullet govuk-list--spaced">
+        {% for offence in courtCase.offences %}
+            <li>
+                {{offence}}
+            </li>
+        {% endfor %}
+    </ul>
+
+    {% if courtCase.sentenceTerms.length > 0 %}
+        <h3>Sentences</h3>
```
davidmiddletonmoj, line 26: Is this the same level of information as the above `<h3>Offences</h3>` or should it be `h4`? (And then `h5` for the following subsections)
Same for the `else` part.

davidmiddletonmoj (approved):
Approved with some minor accessibility comments on heading levels and possible duplication of `formatCurrency` helper

server/utils/utils.ts
jamesgrant-moj: Yes the currency helper is required. It matches the format of the existing profile when values are 4 figures and above. Here is an example on dev - /prisoner/G6123VU/sentence-and-release scroll down to sentence 4

jamesgrant-moj:
Thanks for taking a look @davidmiddletonmoj, I have updated the PR

danielburnley (approved):
LGTM - smol comment on something that I'm not sure is intentionally left, but go ahead and merge it it's all good 🎉

PR #96, opened 2023-04-19
Update care needs to pull data from different endpoints

danielburnley:


server/data/prisonApiClient.ts
```diff
@@ -207,12 +207,15 @@ export default class PrisonApiRestClient implements PrisonApiClient {
   }
 
   async getReferenceCodesByDomain(domain: ReferenceCodeDomain): Promise<ReferenceCode[]> {
-    return this.get<ReferenceCode[]>({ path: `/api/reference-domains/domains/${domain}` })
+    return this.get<ReferenceCode[]>({
```
davidmiddletonmoj, line 210: There is another endpoint (`api/reference-domains/domains/:domain/codes`) which returns a full, unpaged list of all values if you need that. The data structure returned is slightly smaller as well, so as long as you don't need the extra properties, might be worth changing.

davidmiddletonmoj (approved):
Approved with comment about possible endpoint change for performance

server/data/prisonApiClient.ts
danielburnley: ah yeah that might be worth it in the future - if i get a chance at some point ill give it a poke

PR #117, opened 2023-05-11
CDPS-151: Add role based permissions to Update Alert link

davidmiddletonmoj:


server/controllers/alertsController.ts
```diff
@@ -34,8 +36,14 @@ export default class AlertsController {
     // Get prisoner data for banner and for use in alerts generation
     const prisonerData = await this.prisonerSearchService.getPrisonerDetails(req.params.prisonerNumber)
 
+    // Set role based permissions
+    const { authorities: roles = [] } = jwtDecode(res.locals.user.token) as { authorities?: string[] }
+    const canUpdateAlert =
+      roles.some(role => role === Role.UpdateAlert) &&
+      (res.locals.user.activeCaseLoadId === prisonerData.prisonId || ['OUT', 'TRN'].includes(prisonerData.prisonId))
```
danielburnley, line 43: Hmmm - The API when dealing with these two prison (/agency) IDs calls out a particular role (`INACTIVE_BOOKINGS`) - https://github.com/ministryofjustice/prison-api/blob/main/src/main/java/uk/gov/justice/hmpps/prison/service/BookingService.java#L541-L543

Is that something we might want to care about here? I'm not sure

danielburnley, line 43: I wonder if these three lines could be made into a function that takes a role as an argument and returns true/false

davidmiddletonmoj: I thought about that, but the next tickets on Case Notes use slightly different logic, so there'd be no reuse (at least for now). If a few more tickets use the exact logic, then would be worth doing.

davidmiddletonmoj: Good question - I don't know in this case. The Case Notes roles logic does look for that role and those IDs, so guessing it's fine here as-is, but we'll see if anything comes out of testing with more realistic cases/roles.

danielburnley: Ah cool - yeah lets see how the rest shake out at least

danielburnley (approved):

PR #120, opened 2023-05-18
CDPS-147: Case Notes Mop Up

davidmiddletonmoj:
CDPS-150, CDPS-152, CDPS-153, CDPS-163

### DIFF
--- a/architecture-decision-record/0004-use-user-token-for-case-notes-api.md
+++ b/architecture-decision-record/0004-use-user-token-for-case-notes-api.md
@@ -1,0 +1,25 @@
+# 3. Use `async`/`await` pattern over `Promise.then`
+
+Date: 2023-05-15
+
+## Status
+
+✅ Accepted
+
+## Context
+
+Normally we use the `clientToken` to access API endpoints and this is configured with all the roles necessary to use the application.
+
+However, for Case Notes, this API uses some logic to send back sensitive case notes based on role. Giving the client credentials this 
+role would mean *all* users get sensitive case notes but the UI would need to filter them out if the user didn't have the role.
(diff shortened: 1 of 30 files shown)


architecture-decision-record/0004-use-user-token-for-case-notes-api.md
```diff
@@ -0,0 +1,25 @@
+# 3. Use `async`/`await` pattern over `Promise.then`
```
danielburnley, line 1: 👀 need to update the title

server/utils/roleHelpers.ts
```diff
@@ -0,0 +1,9 @@
+import { Role } from '../data/enums/role'
+
+// eslint-disable-next-line import/prefer-default-export
+export const canViewOrAddCaseNotes = (roles: string[], activeCaseLoadId: string, prisonId: string) => {
+  return (
+    (roles?.some(role => role === Role.InactiveBookings) && ['OUT', 'TRN'].includes(prisonId)) ||
+    activeCaseLoadId === prisonId
+  )
+}
```
danielburnley, line 9: Just to confirm that the intended behaviour is as long as it's your active case load and not "a case load that you have"? I'm assuming it's right but just double checking :D

danielburnley (approved):
Just a couple questions but once you've updated that title in the ADR feel free to merge 🎉

server/controllers/overviewController.ts
```diff
@@ -0,0 +1,63 @@
+import { Request, Response } from 'express'
```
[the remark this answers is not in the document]
davidmiddletonmoj: Yeah... routes had no tests - just transposed the logic to the controller to make things easier to read with the added role stuff.
Will create a task to do testing later on.

PR #123, opened 2023-05-22
CDPS-162 - Overview role and caseload checks

danielburnley:


server/utils/utils.ts
davidmiddletonmoj, line 1: In my branch I've created a `Role` enum containing things like `GlobalSearch = 'ROLE_GLOBAL_SEARCH'` which saves typos and is used to check against the roles on `User` without the need to do any string manipulation.
Might be worth refactoring once my PR goes in?

davidmiddletonmoj (approved):
LGTM
Left a comment about future refactor for a `Role` enum coming in my PR

server/utils/utils.ts
danielburnley: Yeah I was wondering that - It'd be good in the JS side but I'm not 100% it would work in the frontend checks at least, though I will give it a go there might be a way 🤔

PR #129, opened 2023-05-30
Check permissions to view prisoners page

danielburnley:


server/routes/index.test.ts
davidmiddletonmoj, line 1: Agree we probably don't need this now.

server/services/overviewPage.test.ts
davidmiddletonmoj, line 1: Again, agree we don't need this

davidmiddletonmoj (approved):
LGTM
Couple of comments agreeing we can remove some tests

PR #134, opened 2023-06-02
CDPS-176 - completed active punishments page

jamesgrant-moj:


server/views/pages/activePunishments.njk
```diff
@@ -0,0 +1,38 @@
+{% extends "./index.njk" %}
+{% from "../macros/conditionallyShow.njk" import conditionallyShow %}
+{% block body %}
+    <div class="govuk-grid-row">
+        <div class="govuk-grid-column-three-quarters">
+            <h1>{{name}}’ active punishments</h1>
```
davidmiddletonmoj, line 6: Just adding the apostrophe only works for names ending in "S"
Needs a new filter to add either `'` or `'s` depending on the last letter of the surname.
Or maybe a new `option` on `formatName`?

server/views/pages/activePunishments.njk
davidmiddletonmoj, line 1: Needs a breadcrumb to go back to the profile otherwise the user is stuck here.
Probably need Dom & Eddie to confirm how that should look as the Figma for this screen only shows `Home > Jones, David` but the Profile has `Back to search results` - do we need to see both here? Does the Profile also need `Home`?
Where does `Home` navigate to?

server/routes/index.ts
```diff
@@ -158,6 +159,27 @@ export default function routes(service: Services): Router {
...
+        ...mapHeaderData(prisonerData, res.locals.user, 'active-punishments', true),
+        ...activePunishmentsPageData,
+        activeTab: false,
+      })
+    })
+  })
+
+  get('/prisoner/:prisonerNumber/adjudications', async (req, res, next) => {
+    checkPrisonerInCaseLoad(req, res, async prisonerData => {
+      res.redirect(`${config.apis.dpsHomePageUrl}/prisoner/${prisonerData.prisonerNumber}/adjudications`)
```
davidmiddletonmoj, line 179: Would be better to use the new `config.serviceUrls.digitalPrison` property

server/views/pages/activePunishments.njk
```diff
@@ -0,0 +1,38 @@
...
+        </div>
+    </div>
+
+    <dl class="govuk-summary-list">
```
davidmiddletonmoj, line 10: I don't think `<dl>` is the right element for this - seems like it would confuse accessibility tools as the content is meant to be a list of `<dt>` followed by `<dd>` tags - not sure this will know which `<dd>` belongs to which `<dt>`?
I think just a table would be better as it is tabular data.

server/views/pages/activePunishments.njk
```diff
@@ -0,0 +1,38 @@
...
+            Details
+        </dt>
+        <dd class="govuk-summary-list__actions govuk-!-font-weight-bold govuk-!-text-align-left hmpps-active-punishments-start-date" data-qa="active-punishment-start-date-label">
+            Start date
+        </dd>
+    </div>
+    {% for award in adjudications.awards %}
+        <div class="govuk-summary-list__row">
+            <dd class="govuk-summary-list__value">
+                {{award.days}} days {{award.sanctionCodeDescription}}
```
davidmiddletonmoj, line 25: Should use the `pluralise` filter here, e.g. `{{ award.days | pluralise('day') }}`

davidmiddletonmoj (commented):
A few comments included. Also, the link to this new page from the adjudications mini card is missing, so there's no way to jump to this page other than manually.
`server/services/overviewPageService.ts:253`

jamesgrant-moj:
Thanks for the feedback @davidmiddletonmoj - ready for another review / approval

davidmiddletonmoj (approved):
LGTM

PR #139, opened 2023-06-06
CDPS-128: Prisoner profile personal page mop up

davidmiddletonmoj:


server/data/localMockData/identifiersMock.ts
```diff
@@ -0,0 +1,59 @@
...
+    type: 'DL',
+    value: 'ABCD/123456/AB9DE',
+    offenderNo: 'G6123VU',
+    issuedDate: '2016-09-08',
+    caseloadType: 'INST',
+    whenCreated: '2016-09-08T09:15:32.160802',
+  },
+  {
+    type: 'NINO',
+    value: 'AB123456A',
```
danielburnley, line 45: Is it worth using the fake one here? - `QQ123456C`

not sure how important it is that we dont use a real one 🤷🏻 but that's the fake one, AB123456A might actually belong to someone (I know AB123456C does)

danielburnley (approved):
LGTM - minor comment but im not sure it particularly matters anyway 😅

server/data/localMockData/identifiersMock.ts
davidmiddletonmoj: Just copied and pasted from other mock data and assumed it was fake. Best to change to a known fake one - updating now.

danielburnley (approved):

PR #140, opened 2023-06-06
Add logic for displaying the most recent address for contacts

danielburnley:
This logic is based on the logic found here: 

https://github.com/ministryofjustice/digital-prison-services/blob/main/backend/controllers/prisonerProfile/personalViewModels/activeContactsViewModel.ts#L9-L20


davidmiddletonmoj (approved):
LGTM

PR #141, opened 2023-06-06
CDPS-187 - Pull aliases from the prison API

danielburnley:
This is to match the existing profile service


davidmiddletonmoj (approved):
LGTM

danielburnley:
Closed as we're leaving it as is because the existing behaviour matches data inside nomis

PR #146, opened 2023-06-08
CDPS-174: Create feedback banner

davidmiddletonmoj:


server/config.ts
```diff
@@ -162,4 +162,5 @@ export default {
   },
   domain: get('INGRESS_URL', 'http://localhost:3000', requiredInProduction),
   localMockData: get('LOCAL_MOCK_DATA', false),
+  feedbackEnabledPrisons: get('FEEDBACK_ENABLED_PRISONS', [], requiredInProduction),
```
danielburnley, line 165: should the default for this not be an empty string rather than an array?

PR #150, opened 2023-06-08
CDPS-182 - fix for offence page duplication

jamesgrant-moj:


server/services/offencesPageService.ts
```diff
@@ -186,11 +192,13 @@ export default class OffencesPageService {
...
-    } else if (courtCase.court && courtCase.court.description) {
+      return courtName
+    }
+    if (courtCase.court && courtCase.court.description) {
       courtName = courtCase.court.description
-    } else {
-      courtName = undefined
+      return courtName
     }
+    courtName = undefined
```
brightonsbox, line 201: Could get rid of the `let courtName` temporary variable and just return `courtCase.agency.description` etc in one line

davidmiddletonmoj (approved):
Couple of comments about typing of `caseId` but approved as it might be fine as-is

server/services/offencesPageService.ts
jamesgrant-moj: Thanks Jon, I'll raise another PR for that change later on.

server/interfaces/prisonApi/sentenceSummary.ts
```diff
@@ -110,4 +118,7 @@ export interface SentenceSummaryTermDetail {
   months?: number
   years?: number
   lifeSentence: boolean
+  licence?: Licence
+  caseId?: string
```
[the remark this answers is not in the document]
jamesgrant-moj: Thanks for taking a look @davidmiddletonmoj . Yes the code base has some inconsistent type definitions for caseIds and bookingIds. I updated some booking ids from strings to numbers in a previous PR. I attempted to switch all case ids in this PR to numbers but I ran into some issues. But I did not want to hold up the bugfix. I will take another look later on and raise another PR if possible.

PR #155, opened 2023-06-09
CDPS-199: No longer caching user caseloads

brightonsbox:


server/services/userService.ts
```diff
@@ -17,12 +16,14 @@ export default class UserService {
   constructor(private readonly hmppsAuthClient: HmppsAuthClient) {}
 
   async getUser(token: string): Promise<UserDetails> {
-    const [user, locations, caseLoads] = await Promise.all([
+    const [user, locations] = await Promise.all([
       this.hmppsAuthClient.getUser(token),
       new PrisonApiClient(token).getUserLocations(),
```
davidmiddletonmoj, line 21: Not sure what the locations are used for, but these will also be cached after switching caseloads - so the locations will appear to be from the previous prison, not the new one

brightonsbox: I've now removed this since it wasn't being used anywhere

davidmiddletonmoj (approved):
LGTM

PR #170, opened 2023-06-13
Increasing number of pods in production to 4

brightonsbox:
4 is the standard number of pods we use in production.


jamesgrant-moj (approved):
LGTM

PR #173, opened 2023-06-13
CDPS-144 - Add X-rays to security section

MeliaBurnley:


davidmiddletonmoj (dismissed):
LGTM, other than I can't see anywhere that does the last bullet from the ticket:
> When the total number of scans reaches 116, show the red warning message in the Figma file (and attached). This should be removed at the start of the next calendar year when the counter will reset to zero.

Not sure if you're doing this under a different PR or will update this one?

davidmiddletonmoj (approved):
LGTM

PR #179, opened 2023-06-14
CDPS-192 - to remove full name

jamesgrant-moj:


server/utils/utils.ts
```diff
@@ -162,6 +162,14 @@ export const formatName = (
   } else if (options?.style === NameFormatStyle.lastCommaFirst) {
     names.unshift(`${names.pop()},`)
     names.pop() // Remove middleNames
+  } else if (options?.style === NameFormatStyle.firstLast) {
```
davidmiddletonmoj, line 165: Could do with extra unit testing for this. Not sure this will handle the situation where there is no middle name(s) to start with?

jamesgrant-moj: Thanks for taking a look. I'll take a closer look at that use case and write an extra unit test.

jamesgrant-moj: @davidmiddletonmoj, I added a unit test for that use case. Please review / approve.

server/utils/utils.test.ts
```diff
@@ -211,6 +211,14 @@ describe('format name', () => {
       'Smith, John',
     ],
     ['All names (LastCommaFirst)', 'John', 'James', 'Smith', { style: NameFormatStyle.lastCommaFirst }, 'Smith, John'],
+    [
```
davidmiddletonmoj, line 214: I think you need another test with no middle name, e.g.
```
[ 
'First name and last name with no middle name (FirstLast)',
       'John',
       undefined,
       'Smith',
       { style: NameFormatStyle.firstLast },
       'John Smith',
     ]
```

server/utils/utils.ts
davidmiddletonmoj: I think this will still not work as expected with no middle name - the first line of the function creates an array based on params passed in and will *always* have a length of 3, but the 2nd element would be `undefined` if no middle names were passed in.
Probably easiest just to remove only the middle name element from the array...
```
} else if (options?.style === NameFormatStyle.firstLast) {
    names.splice(1, 1)
}
```

server/utils/utils.test.ts
jamesgrant-moj: I set the unit test middle name to undefined and the test still passes and the app works. I think my solution is the opposite way to achieve your solution but the disadvantage is there is more lines of code.

} else if (options?.style === NameFormatStyle.firstLast) {
    names.splice(1, 1)
}

Thanks for providing your code example, I have added your solution and it works fine.

jamesgrant-moj:
@davidmiddletonmoj, ready for review / approval.

davidmiddletonmoj (approved):
LGTM

PR #206, opened 2023-06-30
Update Dependcies

danielburnley:
This skips the Cypress update as there were issues with upgrading there due to a bug in cypress


package.json
```diff
@@ -136,45 +136,45 @@
     "@types/csurf": "^1.11.2",
     "@types/express-session": "^1.17.7",
     "@types/http-errors": "^2.0.1",
-    "@types/jest": "^29.5.1",
-    "@types/jsonwebtoken": "^9.0.1",
-    "@types/node": "^18.16.0",
-    "@types/nunjucks": "^3.2.2",
+    "@types/jest": "^29.5.2",
+    "@types/jsonwebtoken": "^9.0.2",
+    "@types/node": "^20.3.2",
```
davidmiddletonmoj, line 141: We're still using node 18.x so better to stick with those types

davidmiddletonmoj (commented):
One minor comment on node types version

package.json
danielburnley: ah nice spot ill put that back to what it was

davidmiddletonmoj (approved):
LGTM

PR #208, opened 2023-07-03
CDPS-226 - render 404 page

jamesgrant-moj:


server/routes/index.ts
```diff
@@ -75,132 +88,156 @@ export default function routes(services: Services): Router {
...
   get('/prisoner/:prisonerNumber', async (req, res, next) => {
-    checkPrisonerInCaseLoad(req, res, async prisonerData => {
-      const overviewController = new OverviewController(
-        services.prisonerSearchService,
-        services.overviewPageService,
-        services.dataAccess.pathfinderApiClientBuilder,
-        services.dataAccess.manageSocCasesApiClientBuilder,
-      )
-      return overviewController.displayOverview(req, res, prisonerData)
+    await checkPrisonerExists(req, res, async () => {
```
davidmiddletonmoj, line 91: I think we need a refactor of `checkPrisonerInCaseLoad` and `checkPrisonerExists` so we're not duplicating API calls, and also change the way we use middleware to make it more flexible for further changes.

Basically, atm `checkPrisonerInCaseLoad` is calling the API endpoint `prisonerSearchClient.getPrisonerDetails()` to get `prisonerData` for it's logic, but then after that `checkPrisonerInCaseLoad` does the exact same thing, so we call the API twice every time.

If we change our `const get = ....` function to take multiple middlewares, we can rewrite each of our route handlers so they are easier to read and extend, something like:
```
get('/prisoner/:prisonerNumber', checkPrisonerExists, checkPrisonerInCaseLoad, async (req, res, next) => {
```
And in the middleware functions themselves, we can pass data from one to the next, so if that data has already been populated from an API call or whatever, the next middleware doesn't need to get it again. Need to be flexible here and check for value(s) being passed through and get if not passed, so people can use middleware independently if needed (i.e. only use `checkPrisonerExists ` but not `checkPrisonerInCaseLoad`).

Not necessarily for this PR, but would also be nice to move logic out of the routing file(s) and into controllers so we keep routing simple.

jamesgrant-moj: @davidmiddletonmoj @whitfield-mj
Sounds good - thanks for taking a look. What changes should be made now for PR approval and what can be a refactor task for later on?

davidmiddletonmoj (approved):
LGTM

PR #212, opened 2023-07-05
CDPS-24: Prisoner profile add a case note page

davidmiddletonmoj:


server/controllers/caseNotesController.ts
```diff
@@ -20,63 +29,332 @@ export default class CaseNotesController {
...
+      if (!canViewCaseNotes(res.locals.user, prisonerData)) {
+        return res.render('notFound.njk', {
+          url: req.headers.referer || `/prisoner/${prisonerData.prisonerNumber}`,
+        })
+      }
+
+      let addCaseNoteLinkUrl: string
+      if (canAddCaseNotes(res.locals.user, prisonerData)) {
+        addCaseNoteLinkUrl = `/prisoner/${prisonerData.prisonerNumber}/add-case-note`
+      }
```
whitfield-mj, line 63:
```suggestion
      const addCaseNoteLinkUrl = canAddCaseNotes(res.locals.user, prisonerData) ? `/prisoner/${prisonerData.prisonerNumber}/add-case-note` : undefined
```

I get nervous when variables are set with let and then mutated, makes me worried about where else they are changed

server/controllers/caseNotesController.ts
```diff
@@ -20,63 +29,332 @@ export default class CaseNotesController {
...
+      const caseNotesUsage = await this.prisonApiClientBuilder(clientToken).getCaseNotesUsage(req.params.prisonerNumber)
+      const hasCaseNotes = Array.isArray(caseNotesUsage) && caseNotesUsage.length
+
+      // Get case notes based on given query params
+      const caseNotesPageData = await this.caseNotesService.get(
+        userToken,
+        prisonerData,
+        queryParams,
+        canDeleteSensitiveCaseNotes,
+      )
```
whitfield-mj, line 75: Can these two be run in parallel seeing as it doesn't look like one depends on the answer of the other?

server/controllers/caseNotesController.ts
```diff
@@ -20,63 +29,332 @@ export default class CaseNotesController {
...
+  public displayAddCaseNote(): RequestHandler {
+    return async (req: Request, res: Response) => {
+      const { clientToken } = res.locals
+      const userToken = res.locals.user.token
+      const { firstName, lastName, prisonerNumber, prisonId, restrictedPatient } =
+        await this.prisonerSearchService.getPrisonerDetails(clientToken, req.params.prisonerNumber)
+      const prisonerDisplayName = formatName(firstName, undefined, lastName, { style: NameFormatStyle.firstLast })
+
+      // If user cannot view this prisoner's case notes, redirect to 404 page
+      if (!canViewCaseNotes(res.locals.user, { prisonId, restrictedPatient } as Prisoner)) {
```
whitfield-mj, line 112: Worrying that this is having to be cast. Perhaps the function arg should receive a partial rather than the full Prisoner if it is not all needed

server/controllers/caseNotesController.ts
```diff
@@ -20,63 +29,332 @@ export default class CaseNotesController {
...
+      const errors = this.validate(type, subType, text, date, hours, minutes)
+      if (!errors.length) {
+        try {
+          await this.caseNotesService.addCaseNote(res.locals.user.token, prisonerNumber, caseNote)
+        } catch (error) {
+          if (error.status === 400) {
+            errors.push({ text: error.message, href: '' })
+          } else throw error
+        }
+      }
```
whitfield-mj, line 179: Why is all this being done in the handler? Isn't there middleware to validate forms?

server/controllers/caseNotesController.ts
```diff
@@ -20,63 +29,332 @@ export default class CaseNotesController {
...
-      ...caseNotesPageData,
-      showingAll,
-      hasCaseNotes,
-      addCaseNoteLinkUrl,
-      staffId,
+    const types = caseNoteTypes
+      ?.filter(t => !onlyActive || t.activeFlag === 'Y')
+      .map(t => ({ value: t.code, text: t.description }))
+    caseNoteTypes.forEach(t => {
+      typeSubTypeMap[t.code] = t.subCodes?.map(s => ({ value: s.code, text: s.description }))
```
whitfield-mj, line 214: I would recommend a reduce rather than instantiating and then mutating 

`const typeSubTypeMap = caseNoteTypes.reduce...`

server/controllers/caseNotesController.ts
davidmiddletonmoj: Just what I'm used to I guess. Seems like a good suggestion to refactor so will do that.

server/controllers/caseNotesController.ts
davidmiddletonmoj: Yeah, fair point. I normally do like a good reduce, but sometimes it hurts my brain.

server/controllers/caseNotesController.ts
whitfield-mj: Well I think when I was here last it was standard, there are certainly examples in other repos. Maybe not the standard approach now?

whitfield-mj: Perhaps it's a new ticket to implement a reusable form validation mechanism. We certainly had a decent example in Managing Intelligence that I'm sure I could dig out

davidmiddletonmoj: No worries... I've done the refactor, just finishing off the unit tests for the new approach.

server/utils/roleHelpers.ts
```diff
@@ -3,10 +3,10 @@ import { userCanEdit, userHasAllRoles } from './utils'
...
 
-export const canViewCaseNotes = (user: User, prisoner: Prisoner) => {
+export const canViewCaseNotes = (user: User, prisoner: Partial<Prisoner>) => {
   return userHasAllRoles([Role.GlobalSearch, Role.PomUser], user.userRoles) || userCanEdit(user, prisoner)
 }
 
-export const canAddCaseNotes = (user: User, prisoner: Prisoner) => {
+export const canAddCaseNotes = (user: User, prisoner: Partial<Prisoner>) => {
   return userHasAllRoles([Role.GlobalSearch, Role.PomUser], user.userRoles) || userCanEdit(user, prisoner)
 }
```
whitfield-mj, line 12: Sorry, think I badly advised here - should have probably been `Pick<Prisoner, 'prisonId' | 'restrictedPatient' ...whatever else it needs>` to ensure type safety

whitfield-mj (approved):

PR #214, opened 2023-07-06
CDPS-230 - update to menu bar style

jamesgrant-moj:


davidmiddletonmoj (commented):
@jamesgrant-moj looks good, but worth checking with Eddie on if "remove rollover state" should mean make it *not* a link?
At the moment it still is tab-able, and has a focus state.
If it shouldn't be a link, `tabLinksMacro.njk` will need updating to turn the active tab into a `<span>` or something - I suspect that was the original intention of the `span` CSS rule just above your changes.

jamesgrant-moj:
@davidmiddletonmoj @danielburnley, I made the tabbing update -  ready for your review / approval.

server/views/partials/profileBanner/macros/tabLinksMacro.njk
```diff
@@ -3,16 +3,24 @@
     {% for tabLink in tabLinks %}
       {% if tabLink.active %}
         <li class="govuk-heading-s hmpps-profile-tab-link-selected" data-qa="active-tab">
+          <span
```
davidmiddletonmoj, line 6: `href` isn't valid for a `<span>` and the link classes don't make any sense if it's not a link, so they can go too and just leave...
```suggestion
          <span data-qa="{{tabLink.label}}">{{ tabLink.label }}</span>
```

assets/scss/components/_header-bar.scss
```diff
@@ -339,6 +339,14 @@
         span {
           color: govuk-colour('black');
         }
+
```
davidmiddletonmoj, line 342: You don't need either of the `a` rules as we're changing it to the `span`, so the above rule will already style that correctly.

jamesgrant-moj: @davidmiddletonmoj  Good spot - updated CSS and classes - ready for review.

davidmiddletonmoj (approved):
LGTM

PR #218, opened 2023-07-10
CDPS-228 - remove CSRA link if prisoner does not belong to user caseload

jamesgrant-moj:


server/mappers/headerMappers.ts
```diff
@@ -15,6 +15,8 @@ export function mapProfileBannerTopLinks(prisonerData: Prisoner, user: User) {
   const { userRoles, caseLoads } = user
   const profileBannerTopLinks = []
 
+  const belongsToCaseLoad = prisonerBelongsToUsersCaseLoad(prisonerData.prisonId, user.caseLoads)
+
   if (prisonerBelongsToUsersCaseLoad(prisonerData.prisonId, caseLoads)) {
```
davidmiddletonmoj, line 20: Should swap use the newly defined `belongsToCaseLoad` in this `if` condition rather than calling the method again.

server/mappers/headerMappers.ts
```diff
@@ -48,7 +50,10 @@ export function mapProfileBannerTopLinks(prisonerData: Prisoner, user: User) {
     hiddenLabel: 'View CSRA history',
     info: prisonerData.csra ? prisonerData.csra : 'Not entered',
     classes: '',
-    url: `${config.serviceUrls.digitalPrison}/prisoner/${prisonerData.prisonerNumber}/csra-history`,
+    url:
+      !belongsToCaseLoad && userHasRoles([Role.GlobalSearch], userRoles)
```
davidmiddletonmoj, line 54: This follows the ticket spec, but not sure this will work as expected as written?
If they *don't* have `GLOBAL_SEARCH` it *will* show the link - maybe don't need the check for global(?) but, what about the other routes to viewing prisoners not in your caseload, e.g. `INACTIVE_BOOKING` (and maybe `userCanEdit`?)
Worth checking with Chris.

davidmiddletonmoj: @jamesgrant-moj Did you manage to catch up with Chris about this?

jamesgrant-moj: Hi @davidmiddletonmoj,
Chris confirmed the logic fir the link should be caseload influenced and INACTIVE_BOOKING and  userCanEdit is not necessary.

"text is always displayed, but the link is only active when the prisoner is in one of the user’s caseloads"

PR #220, opened 2023-07-11
CDPS-257: Remove feedback banner view most recent search from page not found page

davidmiddletonmoj:


server/routes/index.ts
```diff
@@ -42,28 +47,27 @@ export default function routes(services: Services): Router {
...
-            url: req.headers.referer,
-          })
+          return next()
         }
       } else {
-        return res.render('notFound.njk', {
-          url: req.headers.referer,
-        })
+        return next()
       }
```
whitfield-mj, line 54: Could maybe flatten this out make them look more like guard clauses, although i think it is going to be refactored soon anyway

`if (!inactiveBooking) {
  return next()
}

if (!userHasRoles([Role.InactiveBookings], res.locals.user.userRoles)) {
  return next()
}
`

whitfield-mj (approved):

server/routes/index.ts
davidmiddletonmoj: Yeah, we need to move the caseload and exists checks out into middleware - there's a ticket for it already. Can refactor as part of that work.

PR #230, opened 2023-08-02
NON-160 filter inactive and NA in a different prison

Mjwillis:
As NA API endpoint now can filter by current prison and also active only - making use of this to simplify the code


server/services/overviewPageService.test.ts
```diff
@@ -134,17 +134,6 @@ describe('OverviewPageService', () => {
       expect(associationRowTwo[3].text).toEqual('Rival Gang')
     })
 
-    it('Only shows non associations that are part of the same prison', async () => {
```
Mjwillis, line 137: Not needed

danielburnley (approved):

PR #240, opened 2023-08-11
NN-5365: get adjudication data from AdjudicationsAPI instead of PrisonAPI

neilmendum:
https://dsdmoj.atlassian.net/browse/NN-5365

### DIFF
--- a/server/data/adjudicationsApiClient.ts
+++ b/server/data/adjudicationsApiClient.ts
@@ -1,0 +1,23 @@
+import RestClient from './restClient'
+import { AdjudicationsApiClient } from './interfaces/adjudicationsApiClient'
+import { AdjudicationSummary } from '../interfaces/adjudicationSummary'
+import config from '../config'
+
+export default class AdjudicationsApiRestClient implements AdjudicationsApiClient {
+  constructor(private readonly restClient: RestClient) {}
+
+  private async get<T>(args: object, localMockData?: T): Promise<T> {
+    try {
+      return await this.restClient.get<T>(args)
+    } catch (error) {
+      if (config.localMockData === 'true' && localMockData) {
+        return localMockData
+      }
+      return error
+    }
+  }
+
+  async getAdjudications(bookingId: number): Promise<AdjudicationSummary> {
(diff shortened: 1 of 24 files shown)


tests/mocks/prisonApiClientMock.ts
```diff
@@ -7,7 +7,6 @@ export const prisonApiClientMock = (): PrisonApiClient => ({
   getUserCaseLoads: jest.fn(),
   getVisitBalances: jest.fn(),
   getVisitSummary: jest.fn(),
-  getAdjudications: jest.fn(),
```
neilmendum, line 10: Moved to `integration_tests/mockApis/adjudications.ts`

server/data/adjudicationsApiClient.ts
```diff
@@ -0,0 +1,23 @@
...
+    } catch (error) {
+      if (config.localMockData === 'true' && localMockData) {
+        return localMockData
+      }
+      return error
+    }
+  }
+
+  async getAdjudications(bookingId: number): Promise<AdjudicationSummary> {
+    return this.get<AdjudicationSummary>({ path: `/adjudications/by-booking-id/${bookingId}` })
```
neilmendum, line 21: [Here is this endpoint](https://github.com/ministryofjustice/hmpps-manage-adjudications-api/blob/9e4fb75fe356845eda7c96c66aab998f792aecee/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsmanageadjudicationsapi/controllers/AdjudicationSummaryController.kt#L102-L140) in the AdjudicationsAPI

I think we will need to add the `VIEW_ADJUDICATIONS` role to the prisoner profile client credentials as the [AdjudicationSummaryController has it configured](https://github.com/ministryofjustice/hmpps-manage-adjudications-api/blob/9e4fb75fe356845eda7c96c66aab998f792aecee/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsmanageadjudicationsapi/controllers/AdjudicationSummaryController.kt#L29)

davidmiddletonmoj (approved):
LGTM

PR #255, opened 2023-08-21
CDPS-335 neurodiversity section

jamesgrant-moj:


server/views/partials/personalPage/neurodiversity.njk
```diff
@@ -1,16 +1,87 @@
...
+          To check for neurodiversity information for someone in another prison, contact the education team in that
+          prison.
+        </p>
+      {% elseif learnerNeurodivergence.length %}
+        <p class="govuk-body" data-qa="neurodivergence-exists">
+          Information from the local education team - not a medical diagnosis.
+        </p>
+
+        {# Support needed #}
+        {% if learnerNeurodivergence[0].neurodivergenceSupport.length %}
```
davidmiddletonmoj, line 22: Not sure what the actual data looks like, but should this always be hard-coded to the first entry in the array?

Can there be multiple entries? And if so, do they all need to be displayed, or only (latest?) one? What order does the array return entries?

davidmiddletonmoj (commented):
Looks good, but one comment about the array
Also, e2e tests appear broken in pipeline

server/views/partials/personalPage/neurodiversity.njk
jamesgrant-moj: Thanks for taking a look, yes all the curious API responses are arrays which contain single objects.
[https://editor.swagger.io/?url=https://raw.githubusercontent.com/ministryofjustice/curious-API/main/curious-api-specification.yaml](url)

davidmiddletonmoj (approved):
LGTM

PR #263, opened 2023-08-29
CDPS-245: Add alert

davidmiddletonmoj:


server/controllers/alertsController.ts
```diff
@@ -1,62 +1,174 @@
...
+      const alertsPageData = await this.alertsPageService.get(clientToken, prisonerData, queryParams, canUpdateAlert)
+      const showingAll = queryParams.showAll
+
+      // Render page
+      return res.render('pages/alerts/alertsPage', {
+        pageTitle: 'Alerts',
+        ...mapHeaderData(prisonerData, inmateDetail, res.locals.user, 'alerts'),
+        ...alertsPageData,
+        showingAll,
+        addAlertLinkUrl,
```
jamesgrant-moj, line 60: @davidmiddletonmoj, what happens if this is undefined? Does the template handle that?

PR #266, opened 2023-09-01
CDPS-36 - create location details page

jamesgrant-moj:


server/data/prisonApiClient.ts
```diff
@@ -286,6 +290,24 @@ export default class PrisonApiRestClient implements PrisonApiClient {
     return this.get<StaffRole[]>({ path: `/api/staff/${staffId}/${agencyId}/roles` })
   }
 
+  async getAgencyDetails(agencyId: string): Promise<Agencies> {
```
davidmiddletonmoj, line 293: Would it make sense to have `activeOnly` as a method argument, so other callers can set it to true if needed?

server/interfaces/prisonApi/agencies.ts
```diff
@@ -0,0 +1,54 @@
+export interface Agencies {
```
davidmiddletonmoj, line 1: A bit confusing to have these interfaces with plural names when they are just one thing.

server/controllers/prisonerCellHistoryController.ts
```diff
@@ -0,0 +1,168 @@
...
+        const agency = agencyData
+        const agencyDescription = agency ? agency.description : null
+
+        return {
+          establishment: agencyDescription,
+          location: extractLocation(cell.description, agencyName),
+          isTemporaryLocation: isTemporaryLocation(cell.description),
+          movedIn: cell.assignmentDateTime && formatTimestampToDateTime(cell.assignmentDateTime),
+          movedOut: cell.assignmentEndDateTime && formatTimestampToDateTime(cell.assignmentEndDateTime),
+          assignmentDateTime: moment(cell.assignmentDateTime).format('YYYY-MM-DDTHH:mm:ss'),
```
davidmiddletonmoj: Use `date-fns` functions rather than `moment`
But also, we already have some nice date/time helpers in `utils/dateHelpers.ts` which may be useful here. If extra is needed, adding new functions to that file would be nice for future use.

server/controllers/prisonerCellHistoryController.ts
```diff
@@ -0,0 +1,168 @@
...
+          livingUnitId: cell.livingUnitId,
+          agencyId: agencyName,
+          movedInBy: formatName(staffDetails.firstName, '', staffDetails.lastName),
+        }
+      })
+
+      const cellDataLatestFirst = cellData.sort((left, right) =>
+        sortByDateTime(right.assignmentDateTime, left.assignmentDateTime),
+      )
+      const currentLocation = cellDataLatestFirst.slice(0, 1)[0]
```
davidmiddletonmoj, line 114: Don't really need the `slice`, just get first element `cellDataLatestFirst[0]` - it's non-mutating.

server/controllers/prisonerCellHistoryController.ts
```diff
@@ -0,0 +1,168 @@
...
+      const occupants =
+        (currentLocation && (await this.prisonApiClient.getInmatesAtLocation(currentLocation.livingUnitId, {}))) || []
+
+      const previousLocations = cellDataLatestFirst.slice(1)
+
+      const prisonerProfileUrl = `/prisoner/${offenderNo}`
+
+      if (req.query.page) queryParams.page = +req.query.page
```
davidmiddletonmoj, line 122: This block of query params can be removed.

server/controllers/prisonerCellHistoryController.ts
```diff
@@ -0,0 +1,168 @@
...
+      if (req.query.sort) queryParams.sort = req.query.sort as string
+      if (req.query.alertType) queryParams.alertType = req.query.alertType as string[]
+      if (req.query.from) queryParams.from = req.query.from as string
+      if (req.query.to) queryParams.to = req.query.to as string
+      if (req.query.showAll) queryParams.showAll = Boolean(req.query.showAll)
+
+      // Render page
+      return res.render('pages/prisonerCellHistoryPage', {
+        pageTitle: 'Location details',
+        ...mapHeaderData(prisonerData, inmateDetail, res.locals.user, 'location-details'),
```
davidmiddletonmoj, line 132: This is mostly used for the banner on the overview. There's a new method `mapHeaderNoBannerData` for pages where there is no banner, which is more efficient and doesn't require `inmateDetail`, so you might find you don't need that object at all in this controller.

server/utils/utils.ts
```diff
@@ -458,3 +459,64 @@ export const neurodiversityEnabled = (agencyId: string): boolean => {
...
+export const isTemporaryLocation = (locationName: string): boolean => {
+  if (!locationName) return false
+  if (locationName.endsWith('RECP')) return true
+  if (locationName.endsWith('CSWAP')) return true
+  if (locationName.endsWith('COURT')) return true
+  if (locationName.endsWith('TAP')) return true
+  return false
+}
+
+export const formatTimestampToDate = (timestamp: moment.MomentInput, outputFormat = 'DD/MM/YYYY'): moment.MomentInput =>
```
davidmiddletonmoj: Should be using `date-fns` rather than `moment` - also, `dateHelpers.ts` contains some useful functions already, and would be nice to group all date related functions in the same place.

jamesgrant-moj:
Thanks @davidmiddletonmoj for the review - made all changes - ready for review / approval

davidmiddletonmoj (approved):
LGTM

PR #271, opened 2023-09-06
CDPS-271 add alerts and security summary panel to overview

whitfield-mj:


server/interfaces/overviewPage.ts
```diff
@@ -19,9 +19,17 @@ export interface OverviewSchedule {
...
-  text?: string
-}[]
+  nonAssociationName: string
+  offenderNo: string
+  assignedLivingUnitDescription: string
+  reasonDescription: string
+  agencyId: string
+}
+
+export type AlertsSummary = {
```
davidmiddletonmoj, line 29: Any specific reason to use a `type` here rather than `interface`?
We made an ADR to prefer `interface` over `type` unless there's a specific reason... https://github.com/ministryofjustice/hmpps-prisoner-profile/blob/main/architecture-decision-record/0002-prefer-interface-over-type.md

whitfield-mj: No, will change. Just copied the pattern from the one above

davidmiddletonmoj (approved):

PR #283, opened 2023-09-15
Attempt to fix memory leak by removing closures

danielburnley:
Production currently is experiencing a memory leak - as far as I've been able to figure out it was introduced in the release containing this diff: 

https://github.com/ministryofjustice/hmpps-prisoner-profile/compare/7a370b3c1ae4aca696628fdd2620f510e5a501e4...60bfd3f772761e651e9d81665de155ecf8135b50

My initial thoughts are that as a result of us creating a closure which has reference to the services, it doesn't get GC'd as the services themselves are kept in memory. This is a quick change to the code around the controllers that can be deployed and we can monitor it to see if it fixes it. If it doesn't, we can revert this PR


server/routes/csraRouter.ts
```diff
@@ -20,7 +20,7 @@ export default function alertsRouter(services: Services): Router {
     '/prisoner/:prisonerNumber/csra-review',
     getPrisonerData(services, { minimal: true }),
     checkPrisonerInCaseload(),
-    csraController.displayReview(),
+    (req, res, next) => csraController.displayReview(req, res, next),
```
whitfield-mj, line 23: Would this not just be `csraController.displayReview` ?

server/routes/alertsRouter.ts
```diff
@@ -35,21 +35,27 @@ export default function alertsRouter(services: Services): Router {
...
 
   get(
     '/prisoner/:prisonerNumber/add-alert',
     getPrisonerData(services),
     guardMiddleware(GuardOperator.OR, checkHasSomeRoles([Role.UpdateAlert]), checkUserCanEdit()),
-    alertsController.displayAddAlert(),
+    async (req, res, next) => {
+      return alertsController.displayAddAlert(req, res, next)
+    },
   )
```
whitfield-mj, line 59: Don't think you need `async` on the front of these

whitfield-mj (approved):

server/routes/csraRouter.ts
danielburnley: i thought this but it throws a type error 🤷🏻

whitfield-mj (approved):

PR #284, opened 2023-09-15
CDPS-31 schedule page new pr

jamesgrant-moj:


server/controllers/prisonerScheduleController.ts
```diff
@@ -0,0 +1,126 @@
...
+
+    const filterAfternoon = (activities: GetEventScheduleItem[]) =>
+      activities && activities.filter(activity => getHours(parseInt(activity.startTime, 10)) < 17)
+
+    const filterEveningDuties = (activities: GetEventScheduleItem[]) =>
+      activities && activities.filter(activity => getHours(parseInt(activity.startTime, 10)) >= 17)
+
+    const byStartTimeThenByEndTime = (a: GetEventScheduleItem, b: GetEventScheduleItem) => {
+      isBefore(parseInt(a.startTime, 10), parseInt(b.startTime, 10))
+      if (isBefore(parseInt(a.startTime, 10), parseInt(b.startTime, 10))) return -1
```
davidmiddletonmoj, line 70: `a.startTime` looks like it's a full datetime from the mocks?
If so, `parseInt()` isn't going to work properly here.
`isBefore()` takes a `Date` object, so these could be something like `isBefore(new Date(a.startTime), new Date(b.startTime))`

jamesgrant-moj: Thanks for taking a look. The isBefore takes a number or a date.

(alias) isBefore(date: number | Date, dateToCompare: number | Date): boolean
Screenshot 2023-09-18 at 12 18 26

server/controllers/prisonerScheduleController.ts
```diff
@@ -0,0 +1,126 @@
...
+      activities && activities.filter(activity => getHours(parseInt(activity.startTime, 10)) < 12)
+
+    const filterAfternoon = (activities: GetEventScheduleItem[]) =>
+      activities && activities.filter(activity => getHours(parseInt(activity.startTime, 10)) < 17)
+
+    const filterEveningDuties = (activities: GetEventScheduleItem[]) =>
+      activities && activities.filter(activity => getHours(parseInt(activity.startTime, 10)) >= 17)
+
+    const byStartTimeThenByEndTime = (a: GetEventScheduleItem, b: GetEventScheduleItem) => {
+      isBefore(parseInt(a.startTime, 10), parseInt(b.startTime, 10))
```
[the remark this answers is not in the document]
jamesgrant-moj: Good spot

jamesgrant-moj:
Thanks @davidmiddletonmoj - I have updated the PR

server/controllers/prisonerScheduleController.ts
davidmiddletonmoj: Yeah, it takes either, but my concern was the `parseInt` bit - if `a.startTime` is just a numeric string, that's fine then, but if it's an ISO date (which it is in the mocks) like `2023-09-09T12:34:56` then `parseInt` will be "helpful" and just give you the year bit, which is not useful for sorting on the time.
```
parseInt('2023-09-09T12:34:56', 10) // 2023
```

jamesgrant-moj: Ok, that makes sense - I have updated it.

davidmiddletonmoj (approved):
LGTM

PR #293, opened 2023-09-21
CDPS-397 schedule page fixes

jamesgrant-moj:


danielburnley (commented):
It'd be good to add some additional tests around the controller for making sure things are being split up correctly when we've got multiple events across a day - though you only need a couple I think if you reuse some of the code that's in [groupEventsByPeriod](https://github.com/ministryofjustice/hmpps-prisoner-profile/blob/main/server/utils/groupEventsByPeriod.ts) as it already does a lot of this work and is pretty well tested. If you reuse that it should reduce the code in this controller too as it's duplicated between them

jamesgrant-moj:
> It'd be good to add some additional tests around the controller for making sure things are being split up correctly when we've got multiple events across a day - though you only need a couple I think if you reuse some of the code that's in [groupEventsByPeriod](https://github.com/ministryofjustice/hmpps-prisoner-profile/blob/main/server/utils/groupEventsByPeriod.ts) as it already does a lot of this work and is pretty well tested. If you reuse that it should reduce the code in this controller too as it's duplicated between them
> 
> Edit: The types are different between them but as it only uses the `startTime` and `endTime` you can change the type on the util function to allow an object that just has those two fields at least

Thanks for taking a look @danielburnley, I'll update the PR.

server/utils/dateHelpers.ts
```diff
@@ -127,7 +127,7 @@ export const formatDateTime = (
 
 export const timeFormat = (dateTimeStr: string) => {
   const date = new Date(dateTimeStr)
-  const hours = date.getUTCHours()
-  const minutes = date.getUTCMinutes() === 0 ? '00' : date.getUTCMinutes()
+  const hours = date.getHours() < 10 ? `0${date.getHours()}` : date.getHours()
+  const minutes = date.getMinutes() === 0 ? '00' : date.getMinutes()
```
davidmiddletonmoj, line 131: You also need to account for minutes 1 to 9 having a leading 0
Rather than doing these manual conversions, could just replace the whole method content with:
```
return new Date(dateTimeStr).toLocaleTimeString('en-gb', { hour: '2-digit', minute: '2-digit' })
```

server/utils/dateHelpers.test.ts
```diff
@@ -124,7 +124,7 @@ describe('time format', () => {
   it.each([
     ['2023-01-20T12:13:14', '12:13'],
     ['2023-03-23T14:30:14', '14:30'],
-    ['2022-01-10T07:45:14', '7:45'],
+    ['2022-01-10T07:45:14', '07:45'],
     ['2020-11-01T19:10:14', '19:10'],
```
davidmiddletonmoj, line 128: Could add a test with minutes < 10

server/views/pages/prisonerSchedule.njk
davidmiddletonmoj, line 1: Not mentioned on the ticket, but was discussed in the demo.... the print page links should be align right, and the top one should be on the same level as the "View 7 days..." link

davidmiddletonmoj (approved):
LGTM

jamesgrant-moj:
> It'd be good to add some additional tests around the controller for making sure things are being split up correctly when we've got multiple events across a day - though you only need a couple I think if you reuse some of the code that's in [groupEventsByPeriod](https://github.com/ministryofjustice/hmpps-prisoner-profile/blob/main/server/utils/groupEventsByPeriod.ts) as it already does a lot of this work and is pretty well tested. If you reuse that it should reduce the code in this controller too as it's duplicated between them
> 
> Edit: The types are different between them but as it only uses the `startTime` and `endTime` you can change the type on the util function to allow an object that just has those two fields at least

@danielburnley  I attempted to reuse overview logic but ran into issues replicating the behaviour in DPS core. I could not get it to work while reusing the overview logic. So I reverted the code. I did reuse the type ScheduleEvent from the overview code and removed the extra surplus type definition so it is slightly more uniform.

PR #301, opened 2023-09-29
NON-260 Non association count incorrect when user in different caseload

Mjwillis:


server/services/overviewPageService.ts
```diff
@@ -541,9 +541,7 @@ export default class OverviewPageService {
     userCaseloads: CaseLoad[],
   ): AlertsSummary {
     const activeCaseload = userCaseloads.find(caseload => caseload.currentlyActive)
-    const nonAssociationsCount = nonAssociations.filter(
```
Mjwillis, line 544: Filter is not needed the API does this automatically

server/views/partials/overviewPage/alerts/alerts.njk
```diff
@@ -12,7 +12,7 @@
     </div>
     <div class="govuk-summary-list__row--no-border">
         <div class="govuk-!-padding-top-3">
-            <dt class="govuk-!-font-weight-bold govuk-!-margin-bottom-1">Non-associations in {{ user.activeCaseLoad.description }}</dt>
+            <dt class="govuk-!-font-weight-bold govuk-!-margin-bottom-1">Non-associations in {{ prisonerData.prisonName }}</dt>
```
Mjwillis, line 15: Should show where the prisoner is not where the current logged in user active caseload!

whitfield-mj (approved):

PR #311, opened 2023-10-06
CDPS-407 - disable use of force action for 4 prisons

jamesgrant-moj:


server/controllers/overviewController.ts
```diff
@@ -105,7 +105,11 @@ export default class OverviewController {
         dataQA: 'add-key-worker-session-action-link',
       })
     }
-    if (userCanEdit(user, prisonerData) && !prisonerData.restrictedPatient) {
+    if (
+      userCanEdit(user, prisonerData) &&
+      !prisonerData.restrictedPatient &&
+      !config.useOfForceDisabledPrisons.includes(user.activeCaseLoadId)
+    ) {
```
whitfield-mj, line 112: This looks as though it is preventing the add appointment link rather than the use of force one. Is it in the correct place?

whitfield-mj: Would have thought there would be unit tests for stuff like this, it doesn't seem to be well tested in general.

PR #324, opened 2023-10-30
CDPS-280: Add audit service

danielburnley:


server/services/auditService.ts
```diff
@@ -0,0 +1,55 @@
...
+}: {
+  sqsClient: SQSClient
+  queueUrl: string
+  build: string
+  serviceName: string
+}): AuditService => {
+  const sendPageView = async ({ userId, prisonerNumber, details }: PageViewAudit) => {
+    try {
+      const message = JSON.stringify({
+        action: 'PAGE_VIEW',
```
williamfalconeruk, line 36: Might want to be more specific about the page they are viewing if that makes sense?

server/services/auditService.ts
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    const message = JSON.stringify({
+      action: `PAGE_VIEW_${page.toString()}`,
+      when: new Date(),
+      who: userId,
+      subjectId: prisonerNumber,
+      subjectType: 'PRISONER_ID',
+      service: serviceName,
+      requestId, // To be renamed when documentation is updated
```
williamfalconeruk, line 117: which documentation? if it's the HLD then happy to take this as the value.

server/services/auditService.ts
```diff
@@ -0,0 +1,158 @@
...
+
+  const sendAccessAttempt = async ({ userId, userRoles, prisonerNumber, page, requestId }: AccessAttemptAudit) => {
+    const message = JSON.stringify({
+      action: `ACCESS_ATTEMPT_${page.toString()}`,
+      when: new Date(),
+      who: userId,
+      subjectId: prisonerNumber,
+      subjectType: 'PRISONER_ID',
+      service: serviceName,
+      requestId, // To be renamed when documentation is updated
```
williamfalconeruk, line 87: We can add to HLD if this is what's meant by documentation

williamfalconeruk (approved):
happy to approve, but would like the team to ratify this also.

PR #333, opened 2023-11-03
NON-305: Always show links to non-associations in overview page

xoen:
The link to prisoner's non-associations (in the Non-associations service) will be shown always.

Previously we had a short list of prisons that could see these links but we want to show these links all the time now as there is no harm in doing so.

### DIFF
--- a/server/services/overviewPageService.test.ts
+++ b/server/services/overviewPageService.test.ts
@@ -874,7 +874,6 @@ describe('OverviewPageService', () => {
         activeAlertCount: 1,
         nonAssociationsCount: 2,
         nonAssociationsUrl: `${config.serviceUrls.nonAssociations}/prisoner/G6123VU/non-associations`,
-        showNonAssociationsLink: false,
       })
     })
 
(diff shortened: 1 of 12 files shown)


server/services/overviewPageService.test.ts
```diff
@@ -773,43 +772,6 @@ describe('OverviewPageService', () => {
         activeAlertCount: 1,
         nonAssociationsCount: 2,
         nonAssociationsUrl: `${config.serviceUrls.nonAssociations}/prisoner/G6123VU/non-associations`,
-        showNonAssociationsLink: false,
-      })
-    })
-
-    describe('Link to non-associations', () => {
```
xoen, line 780: This test was checking whether we were expecting the non-association link to be shown or not based on the setting with the list of private beta prisons.

But because now the link is always shown this test doesn't make sense anymore (and the integration test now checks for this link presence)

server/services/overviewPageService.ts
```diff
@@ -162,7 +161,7 @@ export default class OverviewPageService {
       offencesOverview,
       prisonName: prisonerData.prisonName,
       staffRoles: staffRoles.map(role => role.role),
-      alertsSummary: this.getAlertsSummary(inmateDetail, nonAssociations, userCaseLoads),
```
xoen, line 165: The user caseloads where used to determine whether to show or nor the link to non-associations. This doesn't happen anymore so this parameter/argument is not necessary anymore.

ushkarev (approved):

ushkarev (approved):

ushkarev (approved):

danielburnley (approved):

PR #338, opened 2023-11-07
CDPS-246: Specific location history page

danielburnley:


server/data/whereaboutsClient.ts
```diff
@@ -12,23 +15,32 @@ export default class WhereaboutsRestApiClient implements WhereaboutsApiClient {
...
-    return this.restClient.post(args)
+  async createAppointments(appointments: AppointmentDefaults): Promise<AppointmentDefaults> {
+    return (await this.restClient.post({ path: `/appointment`, data: appointments })) as AppointmentDefaults
   }
 
-  async getCourts(): Promise<CourtLocation[]> {
-    return this.get<CourtLocation[]>({ path: '/court/courts' })
+  async addVideoLinkBooking(videoLinkBooking: VideoLinkBookingForm): Promise<number> {
+    return (await this.restClient.post({ path: '/court/video-link-bookings', data: videoLinkBooking })) as number
   }
```
whitfield-mj, line 28: Can we not use generics with the post too rather than casting?

server/services/prisonerLocationHistoryService.ts
```diff
@@ -0,0 +1,148 @@
...
+    whereaboutsApi: WhereaboutsApiClient,
+  ) => {
+    try {
+      const cellMoveReason = await whereaboutsApi.getCellMoveReason(bookingId, bedAssignmentHistorySequence)
+      const caseNote = await caseNotesApi.getCaseNote(offenderNo, cellMoveReason.cellMoveReason?.caseNoteId.toString())
+      return caseNote.text
+    } catch (err) {
+      if (err?.status === 404) return null
+      throw err
+    }
```
whitfield-mj, line 50: Does this app not have the ignore 404 thing in the client that @davidmiddletonmoj wrote?

tests/errors/notFoundApiError.ts
```diff
@@ -0,0 +1,8 @@
+export default class NotFoundApiError extends Error {
+  status: number
+
+  constructor() {
+    super()
+    this.status = 404
+  }
+}
```
whitfield-mj, line 8: Do we need a specialist error to throw a 404?

danielburnley: From memory (it's been a bit) - when testing something throwing an API i couldn't create the error that superagent throws, so i had to make my own so you could do `error.status` as it should be

williamfalconeruk (approved):
looks good - i  can't see anything glaringly obvious.

PR #343, opened 2023-11-08
Remove nondtoreleasedate

whitfield-mj:


integration_tests/pages/offencesPage.ts
danielburnley, line 1: probably not the time for it, but it'd be good to not depend on the ordering of thse in the summary lists for the test and use `data-qa` in the future

danielburnley (approved):

integration_tests/pages/offencesPage.ts
whitfield-mj: Yeah I know. Just continued with what was there

PR #361, opened 2023-11-16
Disable modsec rule 

danielburnley:


williamfalconeruk (approved):
Lets see if this resolves the current live issue. We will look to tighten this back up in a separate ticket.

PR #367, opened 2023-11-22
remove address display for POM

whitfield-mj:


server/views/pages/professionalContacts/professionalContactsPage.njk
```diff
@@ -56,18 +56,22 @@
...
-                    </dt>
-                    <dd class="govuk-summary-list__value">
-                        {% if contact.address and not contact.address.noFixedAddress %}
-                            {{ arrayToLines(contact.address | addressToLines, "Not entered") }}
-                        {% else %}
-                            Not entered
-                        {% endif %}
-                    </dd>
-                </div>
+                {% if contact.relationshipDescription !== "Prison Offender Manager"
```
davidmiddletonmoj, line 59: Might be better to use `contact.relationship` which is just a code (e.g. `POM`), just in case the text changes in future?

davidmiddletonmoj (approved):

PR #410, opened 2023-12-15
CDPS-592: Don't display recognised listener unless yes when given no suitible

danielburnley:
Updated based on the logic listed here: 
https://dsdmoj.atlassian.net/browse/NN-2579

### DIFF
--- a/server/services/overviewPageService.ts
+++ b/server/services/overviewPageService.ts
@@ -593,7 +593,14 @@ export default class OverviewPageService {
       })
     }
 
-    if ((recognised === BooleanString.Yes || recognised === BooleanString.No) && suitable !== BooleanString.No) {
+    if (!suitable && recognised === BooleanString.Yes) {
+      statusList.push({
+        label: 'Recognised Listener',
+      })
+    } else if (
+      (recognised === BooleanString.Yes || recognised === BooleanString.No) &&
+      suitable === BooleanString.Yes
+    ) {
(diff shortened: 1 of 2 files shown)


davidmiddletonmoj (approved):

whitfield-mj:
Are there no tests for this logic? Feels like there should be tests

server/services/overviewPageService.ts
```diff
@@ -593,7 +593,14 @@ export default class OverviewPageService {
...
 
-    if ((recognised === BooleanString.Yes || recognised === BooleanString.No) && suitable !== BooleanString.No) {
+    if (!suitable && recognised === BooleanString.Yes) {
+      statusList.push({
+        label: 'Recognised Listener',
+      })
+    } else if (
+      (recognised === BooleanString.Yes || recognised === BooleanString.No) &&
+      suitable === BooleanString.Yes
+    ) {
```
whitfield-mj, line 603: It's blowing my mind. Why would recognised === 'No' mean that it is recognised?!

davidmiddletonmoj:
> Are there no tests for this logic? Feels like there should be tests

There were already tests for everything but the combo that failed. Dan has added that in along with the new logic.

It does seem odd, but that is what is in the specs

danielburnley:
I've updated the original description with the link to where the original logic is, it is very unexpected logic

PR #422, opened 2024-01-11
CDPS-508: CSRA history updates

danielburnley:


server/controllers/csraController.ts
```diff
@@ -35,6 +35,7 @@ export default class CsraController {
       })
 
     const filteredSummaries = this.csraService.filterCsraAssessments(allCsraSummaries, req.query)
+    const filteredAsessments = await this.csraService.getDetailsForAssessments(clientToken, filteredSummaries)
```
brightonsbox, line 38: Is the `csraService.filterCsraAssessments` used elsewhere? Just thinking that maybe these two methods could be condensed into one if we're only ever interested in the details response.

PR #433, opened 2024-01-17
use meta from feComponents instead of env var for activities service

whitfield-mj:


server/utils/isServiceEnabled.ts
```diff
@@ -0,0 +1,5 @@
+import { HeaderFooterMeta } from '../data/interfaces/component'
+
+export default (serviceId: string, componentsMeta: HeaderFooterMeta | undefined): boolean => {
+  return !!componentsMeta?.services?.find(service => service.id === serviceId)
```
timharrison-moj, line 4: Is this only finding whether the `activities` service exists in the metadata, rather than also checking that the current caseload is also in the list?

server/controllers/utils/buildOverviewActions.test.ts
```diff
@@ -74,30 +84,26 @@ describe('buildOverviewActions', () => {
...
-        const user = { ...userMock, userRoles: roles, activeCaseLoadId }
-        const prisoner = { ...PrisonerMockDataA, status: prisonerStatus, prisonId }
+      roles                 | activeCaseLoadId | prisonId | availableServices         | prisonerStatus  | visible
+      ${[Role.ActivityHub]} | ${'MDI'}         | ${'MDI'} | ${[{ id: 'activities' }]} | ${'SOMETHING'}  | ${true}
+      ${[Role.ActivityHub]} | ${'LEI'}         | ${'MDI'} | ${[{ id: 'activities' }]} | ${'SOMETHING'}  | ${false}
+      ${[Role.ActivityHub]} | ${'MDI'}         | ${'LEI'} | ${[{ id: 'activities' }]} | ${'SOMETHING'}  | ${false}
+      ${[Role.ActivityHub]} | ${'MDI'}         | ${'MDI'} | ${[{ id: 'something' }]}  | ${'SOMETHING'}  | ${false}
+      ${[Role.ActivityHub]} | ${'MDI'}         | ${'MDI'} | ${[{ id: 'activities' }]} | ${'ACTIVE OUT'} | ${false}
+      ${[]}                 | ${'MDI'}         | ${'MDI'} | ${[{ id: 'activities' }]} | ${'SOMETHING'}  | ${false}
+    `('user can see: $visible', ({ roles, activeCaseLoadId, prisonId, availableServices, prisonerStatus, visible }) => {
```
timharrison-moj, line 94: Is the test only checking that `activities` is present? Was expecting to see a list of prisons where activities are enabled. The /info endpoints returns ```{
"activeAgencies": [
"LEI",
"RSI",
"PVI",
"STI",
"HEI",
"LPI",
"MDI"
]
}

server/utils/isServiceEnabled.ts
whitfield-mj: The header has already done that work and made the decision if activities should be shown to the user. This is using that decision

server/controllers/utils/buildOverviewActions.test.ts
whitfield-mj: The header has already done that work and made the decision if activities should be shown to the user. This is using that decision

brightonsbox (approved):
LGTM

PR #438, opened 2024-01-22
MOD Security update detection only rules across environments.

williamfalconeruk:
we'll test in dev. if this is ok then we can release to other environments with the same settings.


danielburnley (approved):

helm_deploy/values-dev.yaml
```diff
@@ -7,11 +7,16 @@ generic-service:
     host: prisoner-dev.digital.prison.service.justice.gov.uk
     modsecurity_enabled: true
     modsecurity_audit_enabled: true
-    modsecurity_github_team: "connect-dps"
+    modsecurity_github_team: ""
```
williamfalconeruk, line 10:
```suggestion
    modsecurity_github_team: "connect-dps"
```

whitfield-mj (approved):

PR #443, opened 2024-01-25
CDPS-509: CSRA details updates

danielburnley:


server/interfaces/prisonApi/csraAssessment.ts
```diff
@@ -11,12 +14,15 @@ export interface CsraAssessment {
...
   classificationReviewReason?: string
   nextReviewDate?: string
   offenderNo: string
-  originalClassificationCode?: 'LOW' | 'MED' | 'STANDARD' | 'HI'
+  originalClassificationCode?: ClassificationCode
+  overrideReason?: string
+  overridingClassificationCode?: ClassificationCode
   questions: [
     {
       question: string
```
whitfield-mj, line 28: Does this marry up with the prison api models? Seems like a lot of optional values

whitfield-mj (approved):

server/interfaces/prisonApi/csraAssessment.ts
danielburnley: yeah it matches up - it's a 2-3 stage process so a lot of things can be undefined at the start

PR #445, opened 2024-01-25
CDPS-604: Update breadcrumbs to be consistent

danielburnley:


williamfalconeruk (approved):

server/controllers/prisonerCellHistoryController.ts
```diff
@@ -144,7 +144,7 @@ export default class PrisonerCellHistoryController {
             name: formatName(occupant.firstName, '', occupant.lastName, { style: NameFormatStyle.firstLast }),
             profileUrl: `/prisoner/${occupant.offenderNo}`,
           })),
-        prisonerName: formatName(firstName, '', lastName),
+        prisonerName: formatName(firstName, '', lastName, { style: NameFormatStyle.lastCommaFirst }),
```
davidmiddletonmoj, line 147: Is this the right way round?
This is being set to `lastCommaFirst` but `breadcrumbPrisonerName` below is `firstLast`

danielburnley: oh i didn't notice `breadcrumbPrisonerName` because it's not used in the actual page - that is the right way around though yeah

PR #447, opened 2024-01-26
RR-603 - Add feature toggle for new Work & Skills tab

nathanrussell-moj-digital:
This PR adds a new environment variable and feature toggle for the new Work & Skills tab.

At the moment this PR simply adds the feature toggle. It is not currently used, but our subsequent PRs will start to use it.


mattwills-moj-digital (approved):
LGTM

brightonsbox (approved):
LGTM

We already have a number of feature toggles at the top level of the config.ts export but it's nice to group them like this so our team will need to move existing ones in to stay consistent

PR #451, opened 2024-01-29
RR-575 - Add types to function signatures in workAndSkillsPageService.ts

nathanrussell-moj-digital:
This PR defines the types returned by `workAndSkillsPageService` and adds them to the function signatures.

This is necessary to allow the PLP dev to understand the data returned from `workAndSkillsPageService` to the controller and view, as `workAndSkillsPageService` will be the class that we need to change in order to get Goal data from the PLP service.


server/services/workAndSkillsPageService.ts
```diff
@@ -15,13 +15,41 @@ import { properCaseName } from '../utils/utils'
 import { formatDate } from '../utils/dateHelpers'
 import { RestClientBuilder } from '../data'
 
+interface WorkAndSkillsData {
```
nathanrussell-moj-digital, line 18: `WorkAndSkillsData` is the data type that is returned from the public `get` function in this class (which is what the controller calls)
This definition allows us to "see" / visualise the data, as we will be making changes to this class

server/services/workAndSkillsPageService.ts
```diff
@@ -190,7 +236,10 @@ export default class WorkAndSkillsPageService {
     return govList
   }
 
-  public async getLearnerNeurodivergence(prisonerNumber: string, curiousApiClient: CuriousApiClient) {
+  private async getLearnerNeurodivergence(
```
nathanrussell-moj-digital, line 239: Corrected / strengthened the visibility of this function whilst I was here 👍 (nothing calls it outside of this class, so no need for it to be `public`)

server/services/workAndSkillsPageService.ts
```diff
@@ -157,7 +203,7 @@ export default class WorkAndSkillsPageService {
     return multiListArray
   }
 
-  async getLearnerGoals(prisonerNumber: string, curiousApiClient: CuriousApiClient) {
+  private async getLearnerGoals(prisonerNumber: string, curiousApiClient: CuriousApiClient): Promise<LearnerGoalsData> {
```
nathanrussell-moj-digital, line 206: Corrected / strengthened the visibility of this function whilst I was here 👍 (nothing calls it outside of this class, so no need for it to be effectively `public`)

server/services/workAndSkillsPageService.ts
nathanrussell-moj-digital: Also, I've deliberately not `export`ed them as I'm not sure there's a use case (yet?) to work with these types outside of this class
Defining them in this way was more about having them defined so we have a degree of type safety, and allowing developers to discover and understand the data types used in this class.

Having said that, I'm more than happy to lift these out into the sibling `interfaces` folder and `export` them, if the connect DPS dev team would prefer it that way?

chrimesdev (approved):

PR #456, opened 2024-01-30
RR-575 - Add new REST API client for the Education And Work Plan API (PLP API)

nathanrussell-moj-digital:
This PR adds the new REST API client for the Education And Work Plan API (PLP API)

Functionally it is equivalent to what we've done in PLP UI and CIAG UI, but the patterns are slightly different. I've adopted and followed the patterns already established in this code base, rather than trying to do something new. Specifically:

* I have not imported the swagger spec to generate the types. Instead I have manually implemented them in `server/interfaces/educationAndWorkPlanApi` which is consistent with how/where all the other API response types are defined.
* I have defined and implemented an interface for the REST API client class. We've not bothered with that in PLP & CIAG, but all the other client classes in this code base have one, so I've followed suit.
* What we would know and love as a "test data builder" is not in the `testSupport` directory, but instead is in `server/data/localMockData`. But its the same idea; just in a different directory 🤷‍♂️ 

Other than that, this should look pretty familiar with what we've done in PLP UI and CIAG UI in respect of REST API client classes.

At the moment it's not used by anything. There is no service code calling this, so whilst the URLs in the helm deployments are setup correctly, I've not setup the secrets yet for the system token to have the correct role. That will come soon 👍 

But as it stands there is no risk that the PLP API will be called because there is nothing that is calling/using this new REST API class.

### DIFF
--- a/helm_deploy/values-dev.yaml
+++ b/helm_deploy/values-dev.yaml
@@ -55,6 +55,7 @@ generic-service:
     PRISONER_SEARCH_API_URL: "https://prisoner-search-dev.prison.service.justice.gov.uk"
     PRISON_API_URL: "https://prison-api-dev.prison.service.justice.gov.uk"
     WHEREABOUTS_API_URL: "https://whereabouts-api-dev.service.justice.gov.uk"
+    EDUCATION_AND_WORK_PLAN_API_URL: "https://learningandworkprogress-api-dev.hmpps.service.justice.gov.uk"
 
     # Feature flags
     AUDIT_ENABLED: "true"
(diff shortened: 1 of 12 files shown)


helm_deploy/values-dev.yaml
```diff
@@ -55,6 +55,7 @@ generic-service:
     PRISONER_SEARCH_API_URL: "https://prisoner-search-dev.prison.service.justice.gov.uk"
     PRISON_API_URL: "https://prison-api-dev.prison.service.justice.gov.uk"
     WHEREABOUTS_API_URL: "https://whereabouts-api-dev.service.justice.gov.uk"
+    EDUCATION_AND_WORK_PLAN_API_URL: "https://learningandworkprogress-api-dev.hmpps.service.justice.gov.uk"
```
nathanrussell-moj-digital, line 58: "Education And Work Plan", "PLP" (Personal Learning Plans), and "Learning And Work Progress" are all one and the same thing, and simply reflects the change of name of our service (PLP) over time

mattwills-moj-digital (approved):
LGTM

brightonsbox (approved):
LGTM

PR #457, opened 2024-01-31
CDPS-373: Introducing location-details page data interfaces

brightonsbox:


server/utils/utils.test.ts
```diff
@@ -48,594 +49,638 @@ import {
...
-    ['Lower case', 'robert', 'Robert'],
-    ['Upper case', 'ROBERT', 'Robert'],
-    ['Mixed case', 'RoBErT', 'Robert'],
-    ['Multiple words', 'RobeRT SMiTH', 'Robert Smith'],
-    ['Leading spaces', '  RobeRT', '  Robert'],
-    ['Trailing spaces', 'RobeRT  ', 'Robert  '],
-    ['Hyphenated', 'Robert-John SmiTH-jONes-WILSON', 'Robert-John Smith-Jones-Wilson'],
-  ])('%s convertToTitleCase(%s, %s)', (_: string, a: string, expected: string) => {
-    expect(convertToTitleCase(a)).toEqual(expected)
+describe('utils', () => {
```
brightonsbox, line 52: Sorry just introduced an additional 'describe' block to make it easier to run all tests within IDE, so GitHub thinks all of this file has changed... I'll put another comment where the actual change has been introduced...

server/utils/utils.test.ts
```diff
@@ -48,594 +49,638 @@ import {
...
-        random: 'random1',
-      },
-      {
-        id: 'id2',
-        desc: 'desc2',
-        random: 'random2',
-      },
-    ]
-    const selectOptions = objectToSelectOptions(data, 'id', 'desc')
+  describe('groupBy', () => {
```
brightonsbox, line 652: Here we go - the actual change.

There was no test for the groupBy function.  There are a couple of edge cases that have therefore fallen through the cracks which I think need attention so I'll try to investigate as part of this ticket.

server/utils/utils.test.ts
```diff
@@ -48,594 +49,638 @@ import {
...
+    const group1Item1 = { itemId: 'group1item1', [groupKey]: group1Key }
+    const group1Item2 = { itemId: 'group1item2', [groupKey]: group1Key }
+    const group2Item1 = { itemId: 'group2item1', [groupKey]: group2Key }
+
+    it('groupBy null list produces falsy response', () => {
+      expect(groupBy(null, groupKey)).toBeNull()
+    })
+
+    it('groupBy empty list produces falsy response', () => {
+      // should this not actually just return an empty object??
```
brightonsbox, line 666: Would perhaps have expected just an empty object instead of a 'false' value

server/utils/utils.test.ts
```diff
@@ -48,594 +49,638 @@ import {
...
+      expect(groupBy([], groupKey)).toBeFalsy()
+    })
+
+    it('groupBy list that does not contain the grouping key should return an empty list', () => {
+      // unexpected behaviour....
+      expect(groupBy([group1Item1], 'notthere')).toEqual({ undefined: [group1Item1] })
```
brightonsbox, line 672: grouped by 'undefined' as a key - probably not ideal behavour

server/utils/utils.test.ts
brightonsbox: Oo useful thank you!

PR #459, opened 2024-01-31
RR-575 - Add new PLP service with implementation for Education & Work Plan API

nathanrussell-moj-digital:
This PR is not massive, but adds some important classes relating to getting PLP Goal data. It adds:

* A couple of abstractions of the data that the PLP API returns - `PersonalLearningPlanActionPlan` and `PersonalLearningPlanGoal`. These are basically view models and are not tied to the API response data
* A mapper function, to map between `ActionPlanResponse` (the API response data) and `PersonalLearningPlanActionPlan` (the above view model)
* A service interface `PersonalLearningPlanService` and a concrete implementation `EducationAndWorkPlanApiPersonalLearningPlanService` that uses the `EducationAndWorkPlanApiClient` (aka. PLP API client) to get the PLP Goal related data

The PLP dev team will be familiar with this approach as this is what we've been doing throughout in the PLP and CIAG UI's that we've worked on to date. But for the benefit of others:

The idea of the view models is that they represent the needs of the screen, rather than the specific data that a given API might return. APIs change over time, and may even be swapped out for something else, so not coupling the controller and/or nunjucks views to the API response data is good thing 👍 

The reason I've added an interface for the service class and method (none of the other service classes in this repo have one) is related to the above (decoupling), but specifically because there is a thought that one day the PLP API might be replaced/migrated to the OnePlan API. When that happens PLP Goals will still be a "thing" - they will still be a business domain/concern, and we'll still need to display them on the Work & Skills tab. It's just that we'll get them from a different REST API .... and the API response data might (probably will) look different.
When that happens, it should be an easy job to write a new REST API client class, an implementation of the service interface (that uses the new OnePlan REST API client class), and a new mapper to map from the "new shape" API data returned from the OncePlan REST API to the `PersonalLearningPlanActionPlan` view model.
This approach means that we have a nice loose coupling between an API implementation and the view concerns. Loose coupling and programming to interfaces FTW 😁 
(@martin-ballhatchet-moj - FYI re: your concern that we did not couple prisoner-profile to the PLP API and could easily swap in the OnePlan API as and when 👍 )

Last point of interest is a field in the view model type `PersonalLearningPlanActionPlan` called `problemRetrievingData`.
This is a basic/simple mechanism (that we've adopted in PLP and CIAG UIs) that allows for the API call to fail but not break the page or user journey. It's really simple - the service class wraps the API call in a `try/catch`. If the API call fails, the `catch` block returns an instance of the `PersonalLearningPlanActionPlan` view model, but with `problemRetrievingData` set to true. Then at the controller or (more likely) nunjucks view layer (neither in this PR) we can write a simple if statement that says "if problemRetrievingData display some message saying try again later or similar"
(@williamfalconeruk - the above paragraph relates to the conversation we had earlier re: error handling of API calls and how not to break the page/journey. I'll send you some screen shots/examples from PLP of how this works in practice)

### DIFF
--- a/server/data/localMockData/actionPlanResponse.ts
+++ b/server/data/localMockData/actionPlanResponse.ts
@@ -2,26 +2,40 @@ import { ActionPlanResponse } from '../../interfaces/educationAndWorkPlanApi/act
 import { GoalResponse } from '../../interfaces/educationAndWorkPlanApi/goalResponse'
 import { StepResponse } from '../../interfaces/educationAndWorkPlanApi/stepResponse'
 
-const aValidActionPlanResponseWithOneGoal = (): ActionPlanResponse => {
+const aValidActionPlanResponse = (options?: {
+  reference?: string
+  prisonNumber?: string
+  goals?: Array<GoalResponse>
+}): ActionPlanResponse => {
   return {
-    reference: 'a20912ab-4dae-4aa4-8bc5-32319da8fceb',
-    prisonNumber: 'A1234BC',
-    goals: [aValidGoalResponse()],
+    reference: options?.reference || 'a20912ab-4dae-4aa4-8bc5-32319da8fceb',
+    prisonNumber: options?.prisonNumber || 'A1234BC',
+    goals: options?.goals || [aValidGoalResponse()],
   }
 }
-const aValidGoalResponse = (): GoalResponse => {
+
+const aValidActionPlanResponseWithOneGoal = (): ActionPlanResponse => {
+  return aValidActionPlanResponse({ goals: [aValidGoalResponse()] })
+}
+
+const aValidGoalResponse = (options?: {
+  reference?: string
+  title?: string
+  createdAt?: string
+  updatedAt?: string
+}): GoalResponse => {
   return {
-    goalReference: 'd38a6c41-13d1-1d05-13c2-24619966119b',
-    title: 'Learn Spanish',
+    goalReference: options?.reference || 'd38a6c41-13d1-1d05-13c2-24619966119b',
+    title: options?.title || 'Learn Spanish',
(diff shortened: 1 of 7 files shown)


server/data/localMockData/actionPlanResponse.ts
```diff
@@ -2,26 +2,40 @@ import { ActionPlanResponse } from '../../interfaces/educationAndWorkPlanApi/act
 import { GoalResponse } from '../../interfaces/educationAndWorkPlanApi/goalResponse'
 import { StepResponse } from '../../interfaces/educationAndWorkPlanApi/stepResponse'
 
-const aValidActionPlanResponseWithOneGoal = (): ActionPlanResponse => {
+const aValidActionPlanResponse = (options?: {
+  reference?: string
```
nathanrussell-moj-digital, line 6: I love an options object on our test data builders - it makes them so flexible in their use 😁 (see examples of their use in the tests below)

server/services/personalLearningPlanService.ts
```diff
@@ -0,0 +1,37 @@
...
+import { RestClientBuilder } from '../data'
+import toPersonalLearningPlanActionPlan from '../interfaces/mappers/personalLearningPlanActionPlanMapper'
+import logger from '../../logger'
+
+/**
+ * Interface defining the Personal Learning Plan service methods.
+ *
+ * Concrete implementations are free to use whatever API or other data source in order to implement the service methods.
+ * The use of this interface and the data type abstractions allow other implementations that use other API data sources
+ * to be swapped in as and when necessary.
```
mattwills-moj-digital, line 12: Great kdoc 👍

mattwills-moj-digital (approved):
LGTM

PR #460, opened 2024-01-31
CDPS-373: Moving location-details to use own router file

brightonsbox:


server/routes/alertsRouter.ts
```diff
@@ -12,20 +11,12 @@ import validationMiddleware from '../middleware/validationMiddleware'
...
-      path,
-      handlers.map(handler => asyncMiddleware(handler)),
-    )
-  const post = (path: string | string[], ...handlers: RequestHandler[]) =>
-    router.post(
-      path,
-      handlers.map(handler => asyncMiddleware(handler)),
-    )
+  const get = getRequest(router)
+  const post = postRequest(router)
```
brightonsbox, line 19: A bit of a tidy up with these get post definitions - if people don't like it then let me know and I'll revert this

whitfield-mj: I think it's a good tidy up, I'm just amazed that we still have to wrap in asyncMiddleware. I'm sure express were talking about removing the need 5 years ago

whitfield-mj (approved):

PR #465, opened 2024-02-02
CDPS-633: Handle January on monthly spends

danielburnley:
As JS Dates are 0 indexed for months, we can't perform the shorthand presence check in the query as 0 will always return false.

This commit replaces that shorthand with an explicit check for undefined, falling back to the current month when it is and otherwise getting the month from the query string.


server/controllers/moneyController.ts
```diff
@@ -86,7 +86,7 @@ export default class MoneyController {
     res: Response,
   ) {
     const now = new Date()
-    const month: number = +req.query.month || now.getMonth()
+    const month: number = req.query.month === undefined ? now.getMonth() : +req.query.month
```
whitfield-mj, line 89: could this be `+req.query.month ?? now.getMonth()`

davidmiddletonmoj (approved):

PR #468, opened 2024-02-05
CDPS-373: Fixing appointmentValidator test that fails on a Saturday

brightonsbox:


server/validators/appointmentValidator.test.ts
```diff
@@ -236,7 +236,7 @@ describe('Validation middleware', () => {
   })
 
   it('should fail validation with WEEKDAY repeats starting on a weekend', async () => {
-    const nextWeekendDate = formatDateISO(addDays(new Date(), 6 - new Date().getDay()))
+    const nextWeekendDate = formatDateISO(addDays(new Date(), 6 - new Date().getDay() || 7))
```
brightonsbox, line 239: This change makes sure that Saturdays will get 7 days added to today's date instead of 0

whitfield-mj (approved):

PR #470, opened 2024-02-05
CDPS-373: Moving more location details logic into service

brightonsbox:


server/controllers/prisonerLocationDetailsController.ts
```diff
@@ -69,7 +40,9 @@ export default class PrisonerLocationDetailsController {
         config.featureToggles.moveToReceptionLinkEnabled &&
         canViewCellMoveButton &&
         currentLocation?.location !== 'Reception'
-      const receptionIsFull = !receptions.length
+      const receptionIsFull =
+        canViewMoveToReceptionButton &&
+        (await this.prisonerLocationDetailsPageService.isReceptionFull(clientToken, prisonId))
```
brightonsbox, line 45: Only call the 'getReceptionsWithCapacity` prison API endpoint if necessary

server/controllers/prisonerLocationDetailsController.ts
```diff
@@ -90,7 +63,7 @@ export default class PrisonerLocationDetailsController {
         page: Page.PrisonerCellHistory,
       })
 
-      // TODO CDPS-373: Move into PrisonerLocationDetailsService
+      // TODO CDPS-373: Move into PrisonerLocationDetailsService?
```
brightonsbox, line 66: At risk of opening a can of worms, do we have an agreement with our code structure?

* We have a mix of router definitions defined in `routes/index.ts` and in separate files
* We then have controllers, with varying degrees of logic implemented in the controller, ranging from all of it to almost none.
* We have some services that are aligned to a domain - i.e. `appointmentService.ts`
* We have some services that are purely there to deliver the data for a single page - i.e. `overviewPageService.ts`

How far should a service go? Should it be packaging up all the data needed for a page or should that be the controller's job?

server/controllers/prisonerLocationDetailsController.ts
```diff
@@ -24,43 +24,14 @@ export default class PrisonerLocationDetailsController {
 
     try {
       const { bookingId, firstName, middleNames, lastName, prisonId } = prisonerData
+      const { clientToken } = res.locals
       const name = formatName(firstName, middleNames, lastName, { style: NameFormatStyle.firstLast })
 
       const prisonApiClient = this.prisonApiClientBuilder(res.locals.clientToken)
 
-      const [cells, receptions] = await Promise.all([
```
brightonsbox, line 31: All this gets moved out into the service: `getLocationDetailsByLatestFirst`

server/data/interfaces/prisonApiClient.ts
```diff
@@ -102,7 +102,7 @@ export interface PrisonApiClient {
   getAgencyDetails(agencyId: string): Promise<AgencyLocationDetails>
   getOffenderCellHistory(bookingId: number, params: object): Promise<OffenderCellHistory>
   getReceptionsWithCapacity(agencyId: string, attribute?: string): Promise<Reception[]>
-  getStaffDetails(staffId: string): Promise<StaffDetails>
+  getStaffDetails(username: string): Promise<StaffDetails>
```
brightonsbox, line 105: This was incorrectly named, it's actually the username that's being used to make the API call

server/utils/utils.test.ts
```diff
@@ -382,209 +383,246 @@ describe('utils', () => {
...
+        ],
+        [
+          { html: '<a class="govuk-link govuk-link--no-visited-state" href="/prisoner/DEF321">Guy Incognito</a>' },
+          { text: 'DEF321' },
+          { text: 'NMI-RECP' },
+          { text: 'Rival Gang' },
+        ],
+      ])
     })
   })
```
brightonsbox, line 589: There were some tests that were embedded at the wrong level (grouped with the `getError` tests) so I've moved them out.

server/controllers/prisonerLocationDetailsController.ts
davidmiddletonmoj: @brightonsbox Yeah, it's all evolved as more functionality has been added.

My view is...

Having separate `routes` makes sense where there are multiple associated pages, like case notes (list, add, update), etc. If it's just single pages, not much point moving out of `index`

Controllers should handle view related logic and orchestrate calls to multiple services (where required)

Services should be decoupled from the view, handle "business logic" and orchestrate calls to multiple APIs (where required)

Should probably have a team discussion on this and write an ADR

whitfield-mj: We should def discuss. Def not a fan of the Page services. The controllers should orchestrate calls for data from the services which are domain agnostic. Especially dislike it when they return data shapes for nunjucks. 

Believe that is pretty much what David said anyway!

PR #473, opened 2024-02-06
CDPS-22: Update case note page

davidmiddletonmoj:


server/controllers/caseNotesController.ts
```diff
@@ -211,15 +213,113 @@ export default class CaseNotesController {
...
+      const currentLength = this.calculateCaseNoteTotalLength(currentCaseNote)
+      const isOMICOpen = currentCaseNote.subType === 'OPEN_COMM'
+      const isExternal = Number.isNaN(+currentCaseNote.caseNoteId) // External case notes have non-numeric IDs
+
+      // Prison API adds " ...[%s updated the case notes on yyyy/MM/dd HH:mm:ss] " to the text (%s is variable length username)
+      const maxLength = isExternal
+        ? 4000
+        : 4000 - currentLength - prisonApiAdditionalCaseNoteTextLength - res.locals.user.username.length
+      const caseNoteFlash = req.flash('caseNoteText')
+      const caseNoteText: string = caseNoteFlash?.length ? (caseNoteFlash[0] as never) : ''
```
brightonsbox, line 252: `as string` rather than `as never`?

server/controllers/caseNotesController.ts
```diff
@@ -211,15 +213,113 @@ export default class CaseNotesController {
...
+      const { text, isExternal, currentLength, username } = req.body
+
+      const errors = req.errors || []
+      if (!errors.length) {
+        try {
+          await this.caseNotesService.updateCaseNote(res.locals.user.token, prisonerNumber, caseNoteId, {
+            text,
+            isExternal,
+            currentLength: +currentLength,
+            username,
```
brightonsbox, line 296: This username is provided from the client submitted form so shouldn't be trusted - I think this could be exploited to submit a casenote update as someone else?

server/views/pages/caseNotes/updateCaseNote.njk
```diff
@@ -0,0 +1,105 @@
...
+            {% for amendment in currentCaseNote.amendments %}
+                <div class="govuk-body govuk-!-margin-bottom-2">More details added:</div>
+                <pre class="govuk-body govuk-!-margin-bottom-6" data-qa="more-details-text">{{ amendment.additionalNoteText }}</pre>
+            {% endfor %}
+        {% endif %}
+        <form method="post" novalidate="novalidate" autocomplete="off">
+            <input type="hidden" name="_csrf" value="{{ csrfToken }}" />
+            <input type="hidden" name="isExternal" value="{{ isExternal }}" />
+            <input type="hidden" name="currentLength" value="{{ currentLength }}" />
+            <input type="hidden" name="username" value="{{ user.username }}" />
```
brightonsbox, line 49: Can we avoid the need to pass the user's username like this?

server/controllers/caseNotesController.ts
brightonsbox: As discussed I hadn't understood how username was being used - needed for the form validation and NOT for recording who is submitting the case note update

brightonsbox (approved):
LGTM

PR #481, opened 2024-02-08
RR-584 - Fixes the ordering of PLP goals

nathanrussell-moj-digital:
PR to fix the ordering of PLP goals on the Work & Skills tab


server/interfaces/mappers/personalLearningPlanActionPlanMapper.ts
```diff
@@ -13,7 +13,7 @@ const toPersonalLearningPlanActionPlan = (
   const goals = apiActionPlanResponse.goals.map(goal => toPersonalLearningPlanGoal(goal))
   return {
     prisonerNumber: apiActionPlanResponse.prisonNumber,
-    goals,
+    goals: [...goals],
```
nathanrussell-moj-digital, line 16: Return a copy of the goals array, rather than the goals array itself, because the following line (`getLastUpdatedAuditFields(goals)`) sorts the passed in goals array to find the last updated one.

(ie. its a "pass by reference" vs "pass by value" problem )

chrimesdev (approved):

danielburnley (approved):

PR #482, opened 2024-02-08
RR-584 Put the new PLP Work & Skills page data values behind feature toggle logic

chrimesdev:
## Description

Put the new PLP Work & Skills page data values behind feature toggle logic, so that if the feature toggle is disabled, the old pages still work.

https://dsdmoj.atlassian.net/browse/RR-584


mattwills-moj-digital (approved):
LGTM

PR #500, opened 2024-02-14
CDPS-641: Correcting listener status logic

brightonsbox:


server/services/overviewPageService.test.ts
```diff
@@ -722,10 +722,10 @@ describe('OverviewPageService', () => {
         ['Suitable No/Recognised Blank', suitableListenerNo, recognisedListenerBlank, false, false],
         ['Suitable No/Recognised No', suitableListenerNo, recognisedListenerNo, false, false],
         ['Suitable Yes/Recognised Blank', suitableListenerYes, recognisedListenerBlank, true, false],
-        ['Suitable Yes/Recognised No', suitableListenerYes, recognisedListenerNo, true, true],
+        ['Suitable Yes/Recognised No', suitableListenerYes, recognisedListenerNo, true, false],
```
brightonsbox, line 725: Should only show `Suitable Listener` in this case.  Before it was also showing `Recognised Listener`

davidmiddletonmoj (approved):

PR #502, opened 2024-02-14
CDPS-627: Create probation documents page

danielburnley:


integration_tests/pages/probationDocuments.ts
```diff
@@ -0,0 +1,42 @@
...
+    const details = () => tableRow().find('.govuk-table__cell').eq(0)
+    return {
+      details: () => ({
+        documentLink: () => details().find('.govuk-grid-column-full').eq(0).find('a'),
+        documentTypeAndAuthor: () => details().find('.govuk-grid-column-full').eq(1),
+        documentDescription: () => details().find('.govuk-grid-column-full').eq(2),
+      }),
+      date: () => tableRow().find('.govuk-table__cell').eq(1),
+    }
+  }
```
brightonsbox, line 19: These structures are really nice!

server/routes/index.ts
```diff
@@ -227,6 +228,7 @@ export default function routes(services: Services): Router {
   router.use(professionalContactsRouter(services))
   router.use(goalsRouter(services))
   router.use(locationDetailsRouter(services))
+  router.use(probationDocumentsRouter(services))
```
brightonsbox, line 231: I think David mentioned that routes with only one related endpoint as we have here should be inline in this file, but I do like the idea of organising in separate files, even in this case - maybe worth another quick chat in our catch up on Monday?

brightonsbox (approved):
LGTM!

integration_tests/pages/probationDocuments.ts
danielburnley: creating nice page objects is always a satisfying thing to do :')

PR #523, opened 2024-02-26
CDPS-636: Introducing Result type

brightonsbox:
image-20240301-145053

This PR introduces the `Result` type to neatly handle success / failure in retrieving data from APIs.  It extends from the PromiseSettledResult type that is returned from `Promise.allSettled`, adding some functions to make it easier to handle without `if ... else` blocks everywhere.

It also allows us to convert individual Promises into Results with the `wrap` function, meaning that we can iteratively handle failing results rather than having to deal with them all in one go which for the overview page would be a nightmare!

The `wrap` method also accepts a callback function defined in middleware that sets a `res.locals.pageHasErrors` flag which drives the page error banner at the top of a page.  It is done this way because this could be triggered by any number of APIs failing and it seemed cleaner doing it like this than somehow trying to collect together all the Results and check them all.

### DIFF
--- a/server/services/overviewPageService.test.ts
+++ b/server/services/overviewPageService.test.ts
@@ -183,7 +183,12 @@ describe('OverviewPageService', () => {
   describe('Prison name', () => {
     it('Returns the prison name', async () => {
       const overviewPageService = overviewPageServiceConstruct()
-      const { prisonName } = await overviewPageService.get('token', PrisonerMockDataA, 1, inmateDetailMock)
+      const { prisonName } = await overviewPageService.get({
+        clientToken: 'token',
+        prisonerData: PrisonerMockDataA,
+        staffId: 1,
+        inmateDetail: inmateDetailMock,
+      })
       expect(prisonName).toEqual(PrisonerMockDataA.prisonName)
     })
   })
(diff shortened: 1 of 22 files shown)


server/services/overviewPageService.test.ts
```diff
@@ -802,19 +843,48 @@ describe('OverviewPageService', () => {
...
+        expect(pageErrorFlag).toBeTruthy()
+        expect(
+          res.statuses.some(
+            status =>
+              status.label === 'Support needs unavailable' &&
+              status.subText === 'Try again later' &&
+              status.error === true,
+          ),
+        ).toBeTruthy()
+      })
```
brightonsbox, line 887: This is the only material change to this test

server/services/overviewPageService.ts
```diff
@@ -118,7 +133,7 @@ export default class OverviewPageService {
       prisonerDetail,
     ] = await Promise.all([
       prisonApiClient.getStaffRoles(staffId, userCaseLoads.find(caseload => caseload.currentlyActive)?.caseLoadId),
-      curiousApiClient.getLearnerNeurodivergence(prisonerData.prisonerNumber),
+      Result.wrap(getLearnerNeurodivergence, pageErrorCallback)(),
```
brightonsbox, line 136: When `Promise.all` resolves, `wrap` surrounds the function in a try catch to enable it to return a `Result`.  This avoids the need to turn this into a `Promise.allSettled` and then handle everything immediately.

server/services/overviewPageService.ts
```diff
@@ -607,20 +622,13 @@ export default class OverviewPageService {
...
-            label: 'Support needed',
-            subText: 'Has neurodiversity needs',
-          },
-        ]) ||
-      []
-    )
+  private getNeurodiversitySupportStatus(learnerNeurodivergence: Result<LearnerNeurodivergence[]>): Status[] {
+    const supportNeededStatus = { label: 'Support needed', subText: 'Has neurodiversity needs' }
+    const supportNeededErrorStatus = { label: 'Support needs unavailable', subText: 'Try again later', error: true }
+    return learnerNeurodivergence.handle({
```
brightonsbox, line 628: An example of the `Result.handle` function - hopefully it makes it easier to determine what to do in success / failure cases and the typing forces you to handle both cases.

davidmiddletonmoj (approved):

PR #525, opened 2024-02-27
CDPS-377: Allow setting end date for alerts on creation 

danielburnley:


server/validators/alertValidator.test.ts
```diff
@@ -85,4 +87,28 @@ describe('Validation middleware', () => {
       },
     ])
   })
+
+  fit.each([
```
davidmiddletonmoj, line 91: Should be `it`

danielburnley: oh whoops good spot

davidmiddletonmoj (approved):

PR #526, opened 2024-02-28
CDPS-378 & CDPS-379: Alerts - Add more details & Close

davidmiddletonmoj:


integration_tests/pages/addMoreDetailsPage.ts
```diff
@@ -0,0 +1,25 @@
+import Page, { PageElement } from './page'
+
+export default class AddMoreDetailsPage extends Page {
+  h1 = (): PageElement => cy.get('h1')
+
+  description = (): PageElement => cy.get('[data-qa=alert-description]')
```
danielburnley, line 6: You dont need to change this, but as an awareness thing we do have a `cy.getDataQa` method so you can cut down on the `[data-qa=...]` bits 🎉

danielburnley (approved):

integration_tests/pages/addMoreDetailsPage.ts
davidmiddletonmoj: Oh yeah, forgot about that 👍

PR #536, opened 2024-03-05
CDPS-600: Avoid throwing exceptions when prisoner is not in caseload

davidmiddletonmoj:
I've added `HmppsStatusCode` to the `NotFoundError` so for anything other than the default `NOT_FOUND` status, the `ErrorHandler` now just logs a `WARN` without a stack trace, which doesn't get included in the exceptions alert query due to the existing `where severityLevel >= 3`

Also updated `OverviewPageService` to prevent calling `getStaffRoles` with `undefined` caseload id


brightonsbox (approved):
LGTM

PR #542, opened 2024-03-11
CDPS-681:  remove 'OCG Nominal - Do not share' option from add alerts and remove…

whitfield-mj:
… editing links from active list


server/services/alertsService.ts
```diff
@@ -153,6 +134,40 @@ export default class AlertsService {
...
+        expiredByFullName: formatName(alert.expiredByFirstName, undefined, alert.expiredByLastName),
+      }
+    })
+
+    return {
+      ...pagedAlerts,
+      content: alertContentsForView,
+    }
+  }
+
```
whitfield-mj, line 170: Refactored this a little to enable return of new object rather than mutated data from the api. This was causing a problem with the tests where we were essentially comparing the output to the output, as the mocked input was being changed into the output, so it was erroneously passing

davidmiddletonmoj (approved):

PR #543, opened 2024-03-12
CDPS-680: Display-resettlement-worker

davidmiddletonmoj:


integration_tests/mockApis/prison.ts
```diff
@@ -1206,7 +1206,7 @@ export default {
         headers: {
           'Content-Type': 'application/json;charset=UTF-8',
         },
-        jsonBody: [{ number: '07700000000', type: 'mobile' }],
+        jsonBody: personId === 9876543 ? [] : [{ number: '07700000000', type: 'mobile' }],
       },
     })
   },
```
whitfield-mj, line 1212: These seem incredibly specific. Can we not just have the required response as an argument with a default?

davidmiddletonmoj: Yes, that works better. Thanks.

whitfield-mj (approved):

PR #565, opened 2024-03-27
Switch to use req.middleware to store client token

brightonsbox:


nwilliamsmoj:
I am happy with the intent of the change.

On a more pedantic note would it be better in req.user?

brightonsbox:
> I am happy with the intent of the change.
> 
> On a more pedantic note would it be better in req.user?

I suppose the clientToken is only very loosely related to a user so it feels to me like it might be a bit heavy handed to create a new object within req for this purpose.

davidmiddletonmoj (dismissed):

davidmiddletonmoj (approved):

PR #571, opened 2024-04-02
RR-650 - Added prisonRegisterStore to cache data from prison-register

nathanrussell-moj-digital:
This PR adds `prisonRegisterStore` to cache data from `prison-register`

### DIFF
--- a/server/data/prisonRegisterStore.ts
+++ b/server/data/prisonRegisterStore.ts
@@ -1,0 +1,31 @@
+import { RedisClient } from './redisClient'
+import { PrisonDto } from './interfaces/prisonRegisterApi/prisonRegisterApiTypes'
+import logger from '../../logger'
+
+const ACTIVE_PRISONS = 'activePrisons'
+
+export default class PrisonRegisterStore {
+  constructor(private readonly client: RedisClient) {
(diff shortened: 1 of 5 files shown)


server/data/prisonRegisterStore.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  private async ensureConnected() {
+    if (!this.client.isOpen) {
+      await this.client.connect()
+    }
+  }
+
+  async setActivePrisons(prisons: Array<PrisonDto>, durationDays = 1): Promise<string> {
+    await this.ensureConnected()
+    const activePrisons = prisons.filter(prison => prison.active === true)
```
leejacobson-moj, line 22: Extremely minor, but wondering if this filter logic belongs in this set method? I think I might expect this method to simply set whatever prisons I passed it as the "active prisons".

leejacobson-moj (dismissed):
Looks good. Approved.

server/data/prisonRegisterStore.ts
nathanrussell-moj-digital: Yes, I think you are right ....

leejacobson-moj (approved):

PR #572, opened 2024-04-02
CDPS-679 + CDPS-684: Addresses page

davidmiddletonmoj:


integration_tests/mockApis/prison.ts
```diff
@@ -276,7 +276,7 @@ export default {
         headers: {
           'Content-Type': 'application/json;charset=UTF-8',
         },
-        jsonBody: mockAddresses,
+        jsonBody: resp !== undefined ? resp : mockAddresses,
       },
     })
   },
```
whitfield-mj, line 282: I think these are neater with default args like [here](https://github.com/ministryofjustice/hmpps-prisoner-profile/blob/f8afab9e9528bc2575689371575708978bf2b87b/integration_tests/mockApis/prison.ts#L1355-L1375), but there are a number of each style atm

server/controllers/addressController.ts
```diff
@@ -0,0 +1,97 @@
...
+        value: { html: address.noFixedAddress ? 'No fixed address' : addressToLines(address).join('<br/>') },
+        classes: 'govuk-summary-list__row--no-border',
+      })
+      addressSummary.push({
+        key: { text: 'Type of address' },
+        value: {
+          html: address.addressUsages.filter(usage => usage.activeFlag).length
+            ? address.addressUsages
+                .filter(usage => usage.activeFlag)
+                .map(usage => usage.addressUsageDescription)
```
brightonsbox, line 74: Is `addressUsageDescription` a free text field? This would be a XSS vulnerability if it is since nunjucks doesn't escape html built up in this way.

brightonsbox (approved):
LGTM, just one comment to check

server/controllers/addressController.ts
```diff
@@ -0,0 +1,97 @@
...
+          },
+        })
+      }
+      if (address.comment) {
+        addressSummary.push({
+          key: { text: 'Comment' },
+          value: { text: address.comment },
+        })
+      }
+    }
```
whitfield-mj, line 93: Feels to me like this sort of stuff would be more at home being called from nunjucks with a ` | toSummaryItems`, but again - I know we currently have both styles

server/controllers/addressController.ts
```diff
@@ -0,0 +1,97 @@
...
+      prisonerNumber,
+      prisonerName: formatName(firstName, '', lastName),
+      breadcrumbPrisonerName: formatName(firstName, '', lastName, { style: NameFormatStyle.lastCommaFirst }),
+    })
+  }
+
+  private mapAddressToSummaryItems(address: Address): { rows: GovSummaryItem[]; startDate: string } {
+    const addressSummary: GovSummaryItem[] = []
+
+    if (address) {
```
whitfield-mj, line 62: If you initially include the guard `if(!address) return []` then you can proceed without the indentation. Can also avoid instantiating an empty array and pushing, can just define an array to return

server/controllers/addressController.ts
```diff
@@ -0,0 +1,97 @@
...
+      addressSummary.push({
+        key: { text: 'Address' },
+        value: { html: address.noFixedAddress ? 'No fixed address' : addressToLines(address).join('<br/>') },
```
brightonsbox, line 65: Same here with the XSS concern - worth testing what happens if an address contains `<script>alert('foo');</script>`?

server/controllers/addressController.ts
```diff
@@ -0,0 +1,97 @@
...
+      mailAddress,
+      otherAddresses,
+      prisonerNumber,
+      prisonerName: formatName(firstName, '', lastName),
+      breadcrumbPrisonerName: formatName(firstName, '', lastName, { style: NameFormatStyle.lastCommaFirst }),
+    })
+  }
+
+  private mapAddressToSummaryItems(address: Address): { rows: GovSummaryItem[]; startDate: string } {
+    const addressSummary: GovSummaryItem[] = []
```
brightonsbox, line 60: Rather than having a mutatable object could we do something like this:

```
if (!address) return  { rows: [], startDate: address?.startDate }

const addressSummary: GovSummaryItem = ({
...
})

const addressUsageSummary: GovSummaryItem = ({
...})

return { rows: [addressSummary, addressUsageSummary], startDate: address?.startDate }
```

whitfield-mj: I agree. Maybe one for our dev chat but I would always advocate following functional ideals in general: pure functions, not mutating data etc. Find it makes for much more terse code with less hard to diagnose bugs

server/controllers/addressController.ts
davidmiddletonmoj: @brightonsbox It comes from reference data, so should be fine.

server/controllers/addressController.ts
```diff
@@ -0,0 +1,97 @@
...
+  constructor(
+    private readonly addressService: AddressService,
+    private readonly auditService: AuditService,
+  ) {}
+
+  public async displayAddresses(req: Request, res: Response) {
+    const { firstName, lastName, prisonerNumber, prisonId } = req.middleware.prisonerData
+    const { clientToken } = res.locals
+
+    const addresses = (await this.addressService.getAddresses(clientToken, prisonerNumber)) ?? []
```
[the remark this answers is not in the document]
davidmiddletonmoj: IntelliJ automatically adds them, assuming based on eslint rules

server/controllers/addressController.ts
davidmiddletonmoj: Have updated this function and moved it out to `utils` and am using as a filter called in the template.

whitfield-mj (approved):

PR #580, opened 2024-04-08
MAP-849 added court name to data sent to Notify

GurnankCheema:
Court name needed in Notify template for BVL. 
Will update template once data is being passed through


brightonsbox (approved):
LGTM

PR #581, opened 2024-04-08
RR-650 - Addition of prisonService

nathanrussell-moj-digital:
This PR adds `prisonService` which will be used to lookup prisons by their prisonID. It uses the `prisonRegisterApiClient` to retrieve data, and the `prisonRegisterStore` to cache it.

The code is based on the same service class in PLP.


leejacobson-moj (approved):
Looks good. Approved.

PR #589, opened 2024-04-12
Allowing roles SocDataManager and SocDataAnalyst to refer to soc

timothy-ness:
This PR adds two new roles to the list allowed to refer to soc


brightonsbox (approved):
LGTM

PR #595, opened 2024-04-18
CDPS-650: move some of page overview requests to appropriate services

whitfield-mj:


server/services/overviewPageService.ts
```diff
@@ -55,7 +44,6 @@ export default class OverviewPageService {
     private readonly allocationManagerApiClientBuilder: RestClientBuilder<AllocationManagerClient>,
     private readonly keyworkerApiClientBuilder: RestClientBuilder<KeyWorkerClient>,
     private readonly incentivesApiClientBuilder: RestClientBuilder<IncentivesApiClient>,
```
davidmiddletonmoj, line 46: Assume this will be removed in next PR as no longer used

davidmiddletonmoj (approved):
Minor comment which you're probably already addressing in the next PR

server/services/overviewPageService.ts
whitfield-mj: Yeah that has come out. Will do a full tidy up at the end too

PR #600, opened 2024-04-22
CDPS-442: Feedback banner

davidmiddletonmoj:


helm_deploy/values-dev.yaml
```diff
@@ -64,6 +64,7 @@ generic-service:
     RESTRICTED_PATIENT_API_URL: "https://restricted-patients-api-dev.hmpps.service.justice.gov.uk"
     PRISON_REGISTER_API_URL: "https://prison-register-dev.hmpps.service.justice.gov.uk"
     CALCULATE_RELEASE_DATES_API_URL: "https://calculate-release-dates-api-dev.hmpps.service.justice.gov.uk"
+    CONTENTFUL_HOST: "https://graphql.contentful.com"
```
brightonsbox, line 67: These can be moved into the `values.yaml` file since they're the same in all environments

brightonsbox: Do we all want to include these new env variables as part of our local .env files? I guess it would just error and not show the banner if they're not included?

If you think we should all have the variables locally, it would also be good to update https://github.com/ministryofjustice/hmpps-prisoner-profile/blob/main/readme/building_and_running.md?plain=1#L89

brightonsbox (dismissed):
LGTM

brightonsbox (approved):
LGTM

PR #606, opened 2024-04-29
CDPS-693: Getting user data from the token

brightonsbox:
The original ticket CDPS-693 was to swap the HMPPS Auth `/user/me` endpoint for the equivalent in Manage Users API.

However given that all we were doing with the results of this was putting it in `req.session`, we were not calling this API every request anyway.  All the data that we wanted is present in the token, so there was actually no need to call an API for this at all.  We were already parsing the token for the user roles so this was one extra step to take things like the user's name and staffId from the token.

This has caused a number of ripples so this PR is huge sorry. But a lot of it is ripping stuff out of tests or tidying up so hopefully not too onerous!


integration_tests/e2e/addAlertPage.cy.ts
```diff
@@ -93,7 +93,6 @@ context('Add Alert Page', () => {
       cy.setupUserAuth({
         roles: [Role.GlobalSearch],
         caseLoads: [{ caseloadFunction: '', caseLoadId: 'ZZZ', currentlyActive: true, description: '', type: '' }],
-        activeCaseLoadId: 'ZZZ',
```
brightonsbox, line 96: Don't need this now, derived from the caseloads list above for which is `currentlyActive`

integration_tests/e2e/login.cy.ts
```diff
@@ -76,13 +76,14 @@ context('SignIn', () => {
...
     cy.visit('/prisoner/G6123VU')
     const indexPage = Page.verifyOnPage(IndexPage)
+
     cy.task('stubVerifyToken', false)
 
     // can't do a visit here as cypress requires only one domain
     cy.request('/prisoner/G6123VU').its('body').should('contain', 'Sign in')
 
     cy.task('stubVerifyToken', true)
-    cy.task('stubAuthUser', { name: 'bobby brown' })
```
brightonsbox, line 85: Instead of stubbing the /user/me call to HMPPS Auth, update the token

integration_tests/mockApis/auth.ts
```diff
@@ -118,41 +128,16 @@ const token = (userRoles: string[]) =>
     },
   })
 
-const stubUser = (name: string, activeCaseLoadId?: string) =>
```
brightonsbox, line 121: We don't make the /user/me HMPPS Auth call any more

server/controllers/addressController.ts
```diff
@@ -28,9 +28,7 @@ export default class AddressController {
 
     this.auditService
       .sendPageView({
-        userId: res.locals.user.username,
-        userCaseLoads: res.locals.user.caseLoads,
-        userRoles: res.locals.user.userRoles,
+        user: res.locals.user,
```
brightonsbox, line 31: We now have a stronger interface (`HmppsUser`) for what is stored in `res.locals.user` which forces compile time checking that the user has the concept of case loads before trying to access them on the object.  It seemed cleaner to pass the whole `HmppsUser` object through to the audit service methods

server/controllers/appointmentsController.test.ts
```diff
@@ -109,25 +120,15 @@ describe('Appointments Controller', () => {
       body: {},
       query: {},
       path: 'appointments',
-      session: {
```
brightonsbox, line 112: We're no longer storing the results of the HMPPS Auth API call in `req.session` since there's minimal overhead in inspecting the token for every request compared to calling an API.

server/data/hmppsAuthClient.ts
```diff
@@ -32,40 +30,6 @@ function getSystemClientTokenFromHmppsAuth(username?: string): Promise<superagen
...
-  activeCaseLoadId: string
-  userRoles: string[]
-  caseLoads: CaseLoad[]
-}
-
-export interface UserRole {
-  roleCode: string
-}
-
-export default class HmppsAuthClient {
```
brightonsbox, line 46: None of this required any more!

server/data/manageUsersApiClient.test.ts
```diff
@@ -0,0 +1,37 @@
+import nock from 'nock'
```
brightonsbox, line 1: When considering swapping HMPPS Auth to the Manage Users API endpoint, I noticed that there was no test for this client

server/routes/testutils/appSetup.ts
```diff
@@ -49,7 +37,7 @@ function appSetup(services: Services, production: boolean, userSupplier: () => E
 export function appWithAllRoutes({
   production = false,
   services = {},
-  userSupplier = () => user,
+  userSupplier = () => null,
```
brightonsbox, line 40: This is just propping up a single test to show that the app responds properly with a not found page - a lot of stuff here isn't actually being used...

server/interfaces/HmppsUser.ts
```diff
@@ -0,0 +1,30 @@
...
+  userRoles: string[]
+  token?: string
+}
+
+export interface PrisonUser extends BaseUser {
+  authSource: 'nomis'
+  staffId: number
+  activeCaseLoadId: string
+  caseLoads: CaseLoad[]
+}
```
brightonsbox, line 20: Hopefully this will force the consideration - is this a Prison user? Do they actually have the concept of a case load or a staffId etc.

server/views/dataUtils/categorySummaryToMiniSummary.ts
```diff
@@ -5,7 +5,7 @@ import OverviewPageData from '../../controllers/interfaces/OverviewPageData'
 
 export default (
   categorySummary: OverviewPageData['categorySummary'],
-  prisonerInCaseload: boolean,
```
brightonsbox, line 8: Correcting small casing typo

whitfield-mj (approved):

PR #607, opened 2024-04-30
CDPS-738: Make update alert use lockTimeout

davidmiddletonmoj:


steverendell:
The original PR also had a part which showed a different error message when it received a 423, do you need to include it?

davidmiddletonmoj:
> The original PR also had a part which showed a different error message when it received a 423, do you need to include it?

Thanks Steve, yes we should include something similar.

whitfield-mj (approved):

PR #610, opened 2024-05-02
CDPS-741: Adding typing to res.locals.user

brightonsbox:


server/controllers/appointmentController.ts
```diff
@@ -52,16 +53,14 @@ export default class AppointmentController {
   public displayAddAppointment(): RequestHandler {
     return async (req: Request, res: Response, next: NextFunction) => {
       const { clientToken } = req.middleware
-      const {
-        user: { activeCaseLoadId },
-      } = res.locals
       const { prisonerNumber } = req.params
       const { firstName, lastName, bookingId, cellLocation, prisonId } = req.middleware.prisonerData
       const prisonerName = formatName(firstName, undefined, lastName, { style: NameFormatStyle.lastCommaFirst })
+      const user = res.locals.user as PrisonUser
```
brightonsbox, line 59: So this is now forcing us to consider that the user could not be a `PrisonUser` and therefore might not have the concept of a `caseload` or a `staffId`.  Might seem a bit annoying, but it's already flushed out a few things such as the question, 'should non-prison users have access to these pages?'

Here I've cast the type to PrisonUser because the check is done in middleware.  I think that's probably the best way to do it, or we could do a check and `return next`? 🤔

server/controllers/caseNotesController.ts
```diff
@@ -81,9 +81,6 @@ export default class CaseNotesController {
       )
       const showingAll = queryParams.showAll
 
-      // Get staffId to use in conditional logic for amend link
-      const { staffId } = res.locals.user
-
```
brightonsbox, line 86: Not needed, nunjucks already has access to res.locals.user

server/controllers/overviewController.ts
```diff
@@ -87,7 +86,7 @@ export default class OverviewController {
       offencesOverview,
       nonAssociationSummary,
     ] = await Promise.all([
-      activeCaseLoadId ? this.userService.getStaffRoles(clientToken, staffId, activeCaseLoadId) : ([] as StaffRole[]),
+      this.userService.getStaffRoles(clientToken, res.locals.user),
```
brightonsbox, line 89: Pass in user, allows `userService.getStaffRoles` to decide whether it needs to do the API call at all

server/middleware/bannerMiddleware.ts
```diff
@@ -6,7 +6,7 @@ import { Services } from '../services'
 export default function bannerMiddleware({ contentfulService }: Services): RequestHandler {
   return async (req, res, next) => {
     try {
-      res.locals.bannerText = await contentfulService.getBanner(res.locals.user.activeCaseLoadId)
+      res.locals.bannerText = await contentfulService.getBanner(res.locals.user)
```
brightonsbox, line 9: again pass in the user object, let the service deal with the user types

server/middleware/populateCurrentUser.ts
```diff
@@ -39,14 +43,13 @@ export function populateCurrentUser(): RequestHandler {
 export function getUserCaseLoads(userService: UserService): RequestHandler {
   return async (req, res, next) => {
     try {
-      if (res.locals.user) {
+      if (res.locals.user && res.locals.user.authSource === 'nomis') {
```
brightonsbox, line 46: Don't need to call the API to retrieve case loads if it's not a prison user

server/routes/appointmentRouter.ts
```diff
@@ -91,12 +91,16 @@ export default function appointmentRouter(services: Services): Router {
   get(
     '/api/get-offender-events',
     auditPageAccessAttempt({ services, page: ApiAction.OffenderEvents }),
+    isCreateIndividualAppointmentRolledOut,
+    checkPrisonerInCaseload({ activeCaseloadOnly: true }),
```
brightonsbox, line 95: These API routes should have the same protection as the page that's calling them

whitfield-mj (approved):

PR #611, opened 2024-05-03
Allowed pathfinder user role to refer to pathfinder from the profile

timothy-ness:
This PR is for allowing users with the role PATHFINDER_USER to refer prisoners to pathfinder.

### DIFF
--- a/integration_tests/e2e/overviewPage.cy.ts
+++ b/integration_tests/e2e/overviewPage.cy.ts
@@ -315,6 +315,27 @@ context('Overview Page', () => {
     })
   })
 
+  context('Given the user has PF_LOCAL_READER role', () => {
+    beforeEach(() => {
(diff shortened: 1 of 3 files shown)


integration_tests/e2e/overviewPage.cy.ts
davidmiddletonmoj, line 1: The `Given the user has PF_USER role` tests were to confirm a user that only had a "read only" type Pathfinder role would not see the "Refer" link, but would see the "Profile" link.
Now that `PF_USER` *does* grant access to "Refer", it's useful to have a test confirm that, but we should still have a test that uses a "read only" type role (e.g. `PF_LOCAL_READER`) and confirms non-display of the "Refer" link.

davidmiddletonmoj (approved):

PR #619, opened 2024-05-15
Bringing closer in line with `hmpps-typescript-template`

brightonsbox:


server/views/partials/breadCrumb.njk
```diff
@@ -1,32 +0,0 @@
-{% from "govuk/components/breadcrumbs/macro.njk" import govukBreadcrumbs %}
```
brightonsbox, line 1: Was not being used.

whitfield-mj (approved):

PR #621, opened 2024-05-16
CDPS-721: Use parameterised paths for prometheus metrics

brightonsbox:


server/data/restClientMetricsMiddleware.test.ts
```diff
@@ -13,20 +13,18 @@ describe('restClientMetricsMiddleware', () => {
   })
 
   describe('normalizePath', () => {
-    it('removes the query params from the URL path', () => {
```
brightonsbox, line 16: Now only the path is passed in rather than the full URI, so don't need to worry about query params

davidmiddletonmoj (approved):

PR #627, opened 2024-05-23
CDPS-746: Alerts API switching part 2

davidmiddletonmoj:


integration_tests/mockApis/alertsApi.ts
```diff
@@ -0,0 +1,258 @@
...
+import { alertDetailsExpiresMock, alertDetailsMock } from '../../server/data/localMockData/alertDetailsMock'
+
+export default {
+  stubActiveAlerts: (prisonerNumber: string) => {
+    let jsonResp
+    if (prisonerNumber === 'G6123VU') {
+      jsonResp = pagedActiveAlertsMock
+    } else if (prisonerNumber === 'A1234BC') {
+      jsonResp = emptyAlertsMock
+    }
```
whitfield-mj, line 20: Would suggest passing in the payloads to the mock rather than conditional based on prisonerNumber as it makes it easier to read what is happening from within the test

integration_tests/mockApis/componentApi.ts
```diff
@@ -0,0 +1,23 @@
...
+      request: {
+        method: 'GET',
+        urlPattern: '/component/.*',
+      },
+      response: {
+        status: 200,
+        headers: {
+          'Content-Type': 'application/json;charset=UTF-8',
+        },
+        jsonBody: resp ?? componentsMock,
```
whitfield-mj, line 19: not going to be undefined if it has a fallback arg

server/controllers/alertsController.ts
```diff
@@ -272,10 +268,13 @@ export default class AlertsController {
 
     // Initialise form
     const alertFlash = req.flash('alert')
-    const { description } = alertFlash?.length ? (alertFlash[0] as never) : { description: alert.description }
+    const { description, activeTo } = alertFlash?.length
+      ? (alertFlash[0] as never)
```
whitfield-mj, line 272: Why would that be as never?

server/data/alertsApiClient.ts
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  async updateAlert(alertId: string, alertChanges: AlertChanges): Promise<Alert> {
+    return this.restClient.put<Alert>({ path: `/alerts/${alertId}`, data: alertChanges })
+  }
+
+  async getAlertTypes(): Promise<AlertType[]> {
+    return this.restClient.get<AlertType[]>({ path: '/alert-types' })
+  }
+}
```
whitfield-mj, line 42: Pedantic but functions don't need to be marked as async if there's no await within

server/services/alertsService.ts
```diff
@@ -44,17 +56,47 @@ export default class AlertsService {
...
-    if (apiParams.from) apiParams.from = apiParams.from && formatDateISO(parseDate(apiParams.from))
-    if (apiParams.to) apiParams.to = apiParams.to && formatDateISO(parseDate(apiParams.to))
-    if (apiParams.page) apiParams.page = apiParams.page && +apiParams.page - 1 // Change page to zero based for API query
+    if (apiParams.from) apiParams.from = formatDateISO(parseDate(apiParams.from))
+    if (apiParams.to) apiParams.to = formatDateISO(parseDate(apiParams.to))
+    if (apiParams.page) apiParams.page = +apiParams.page - 1 // Change page to zero based for API query
+    apiParams.size = queryParams?.showAll ? 9999 : 20
+
+    return apiParams
+  }
```
whitfield-mj, line 73: Would prefer something that doesn't clone and mutate e.g.
```javascript
return {
  from: apiParams.from ? formatDateISO(parseDate(apiParams.from)) : undefined,
  to: apiParams.to ? formatDateISO(parseDate(apiParams.to)) : undefined,
  ...
}
```

server/services/alertsService.ts
```diff
@@ -44,17 +56,47 @@ export default class AlertsService {
...
+      isActive: boolean
+      alertType?: string | string[]
+      alertCode?: string | string[]
+      activeFromStart?: string
+      activeFromEnd?: string
+      search?: string
+      page?: number
+      size: number
+      sort?: string
+    } = { isActive: queryParams.alertStatus === 'ACTIVE', size: queryParams?.showAll ? 9999 : 20 }
```
whitfield-mj, line 92: Nicer to name this type?

server/services/alertsService.ts
```diff
@@ -44,17 +56,47 @@ export default class AlertsService {
...
+
+    if (queryParams.alertType) apiParams.alertType = queryParams.alertType
+    if (queryParams.from) apiParams.activeFromStart = formatDateISO(parseDate(queryParams.from))
+    if (queryParams.to) apiParams.activeFromEnd = formatDateISO(parseDate(queryParams.to))
+    if (queryParams.page) apiParams.page = +queryParams.page - 1 // Change page to zero based for API query
+    if (queryParams.sort)
+      apiParams.sort = queryParams.sort.replace('dateCreated', 'activeFrom').replace('dateExpires', 'activeTo')
```
whitfield-mj, line 99: as above re mutation

server/controllers/alertsController.ts
davidmiddletonmoj: Typescript allows you to use `never` similar to `any` in some situations where you just don't care what the type is, but agree it's more useful to actually describe the type properly so have updated

server/data/alertsApiClient.ts
davidmiddletonmoj: I prefer to still use it tbh, just so it's obvious what is happening.

whitfield-mj: Fair

PR #638, opened 2024-06-05
CDPS-767: Move alertsApiEnabled flag into Redis FeatureToggleStore

davidmiddletonmoj:


server/services/featureToggleService.ts
```diff
@@ -0,0 +1,15 @@
+import FeatureToggleStore from '../data/featureToggleStore/featureToggleStore'
+
+export default class FeatureToggleService {
+  constructor(private readonly featureToggleStore: FeatureToggleStore) {}
+
+  public async getFeatureToggle(prisonId: string, featureToggle: string): Promise<boolean> {
+    if (!prisonId) return false
```
brightonsbox, line 7: I wonder if we were to use this for other things, we'd want to be able to supply a default (might not always be false by default?)

brightonsbox: One to implement when we get to it though, not necessarily for this PR

brightonsbox (approved):
LGTM

server/middleware/frontEndComponents.ts
```diff
@@ -23,7 +25,14 @@ export default function getFrontendComponents({ componentService }: Services, us
       res.locals.feComponentsMeta = meta
 
       /* Set feature toggle for using Alerts API */
-      config.featureToggles.alertsApiEnabled = meta?.services?.some(service => service.id === 'alerts')
+      if ('activeCaseLoadId' in res.locals.user) {
+        await featureToggleService.setFeatureToggle(
```
nwilliamsmoj, line 29: If this does have performance implications, then we could incorporate a scheduled task in to the feature toggle store to go and fetch it every minute or so.

PR #642, opened 2024-06-07
CDPS-773: Add secondary sort parameters to alerts list queries for Alerts API

davidmiddletonmoj:
CDPS-779: Use Alerts API to get counts for tabs in enabled prisons
Add defined properties to `Request.middleware` interface declaration

### DIFF
--- a/server/services/alertsService.ts
+++ b/server/services/alertsService.ts
@@ -15,6 +15,7 @@ import {
   toAlert,
   toAlertsApiCreateAlert,
   toAlertType,
+  toAlertTypesFilters,
   toPrisonAlertChanges,
   toPrisonApiCreateAlert,
 } from './mappers/alertMapper'
@@ -205,28 +206,8 @@ export default class AlertsService {
 
   private async getPrisonApiAlertSummaryData(prisonApiClient: PrisonApiClient, bookingId: number) {
     const { activeAlertCount, inactiveAlertCount, alerts } = await prisonApiClient.getInmateDetail(bookingId)
-    const activeAlertTypesFilter: { [key: string]: AlertTypeFilter } = {}
-    const inactiveAlertTypesFilter: { [key: string]: AlertTypeFilter } = {}
 
-    alerts.forEach(alert => {
-      if (alert.active) {
-        activeAlertTypesFilter[alert.alertType] = {
-          code: alert.alertType,
-          description: alert.alertTypeDescription,
-          count: activeAlertTypesFilter[alert.alertType] ? activeAlertTypesFilter[alert.alertType].count + 1 : 1,
-          checked: false,
-        }
-      } else {
-        inactiveAlertTypesFilter[alert.alertType] = {
-          code: alert.alertType,
-          description: alert.alertTypeDescription,
-          count: inactiveAlertTypesFilter[alert.alertType] ? inactiveAlertTypesFilter[alert.alertType].count + 1 : 1,
-          checked: false,
-        }
-      }
-    })
-
-    return { activeAlertCount, inactiveAlertCount, activeAlertTypesFilter, inactiveAlertTypesFilter }
+    return { activeAlertCount, inactiveAlertCount, ...toAlertTypesFilters(alerts.map(toAlert)) }
   }
 
   /* eslint-disable @typescript-eslint/no-unused-vars */
--- a/server/services/alertsService.ts
+++ b/server/services/alertsService.ts
@@ -18,7 +18,7 @@ import {
   toPrisonAlertChanges,
   toPrisonApiCreateAlert,
 } from './mappers/alertMapper'
-import { Alert, AlertChanges } from '../data/interfaces/alertsApi/Alert'
+import { Alert, AlertChanges, AlertsApiQueryParams, AlertSummaryData } from '../data/interfaces/alertsApi/Alert'
 import { AlertsApiClient } from '../data/interfaces/alertsApi/alertsApiClient'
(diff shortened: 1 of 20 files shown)


server/services/alertsService.ts
```diff
@@ -195,6 +203,32 @@ export default class AlertsService {
...
+        inactiveAlertTypesFilter[alert.alertType] = {
+          code: alert.alertType,
+          description: alert.alertTypeDescription,
+          count: inactiveAlertTypesFilter[alert.alertType] ? inactiveAlertTypesFilter[alert.alertType].count + 1 : 1,
+          checked: false,
+        }
+      }
+    })
+
+    return { activeAlertCount, inactiveAlertCount, activeAlertTypesFilter, inactiveAlertTypesFilter }
```
whitfield-mj, line 229: Think it would be preferable as a reduce to avoid mutation 
Something like....

```javascript
const {active, inactive} = alerts.reduce((output, alert) => {
  if (alert.active) {
    return {
      ...output,
      active: { ...output.active, [alert.alertType]: { ... } }
    }

    return {
      ...output,
      inactive: { ...output.inactive, [alert.alertType]: { ... } }
  }
}, {active: {}, inactive: {}})
```

server/services/mappers/alertMapper.ts
```diff
@@ -140,3 +148,40 @@ export const toAlertFlagLabels = (alerts: Alert[], alertFlags: AlertFlagLabel[])
...
+        count: inactiveAlertTypesFilter[alert.alertCode.alertTypeCode]
+          ? inactiveAlertTypesFilter[alert.alertCode.alertTypeCode].count + 1
+          : 1,
+        checked: false,
+      }
+    }
+  })
+
+  return { activeAlertTypesFilter, inactiveAlertTypesFilter }
+}
```
whitfield-mj, line 178: As above

server/services/alertsService.ts
brightonsbox: Agreed on trying to avoid the pattern of declaring and then mutating objects if possible, maybe worth a discussion about procedural / functional style at our dev catch up since I don't think we've ever set anything in stone before?

brightonsbox: Would something like this work:

```

const toAlertTypeFilters = ({ alerts, active }: { alerts: Alert[]; active: boolean }) => {
  return alerts
    .filter(alert => alert.isActive === active)
    .reduce((alertCounts, alert) => {
      // eslint-disable-next-line no-param-reassign
      alertCounts[alert.alertCode.alertTypeCode] = {
        code: alert.alertCode.alertTypeCode,
        description: alert.alertCode.alertTypeDescription,
        count: (alertCounts[alert.alertCode.alertTypeCode] ?? 0) + 1,
        checked: false,
      }
      return alertCounts
    }, {})
}

export const toAlertSummaryData = (alerts: Alert[], alertFlags: AlertFlagLabel[]): AlertSummaryData => {
  return {
    activeAlertCount: alerts.filter(alert => alert.isActive).length,
    activeAlertTypesFilter: toAlertTypeFilters({ alerts, active: true }),
    inactiveAlertCount: alerts.filter(alert => !alert.isActive).length,
    inactiveAlertTypesFilter: toAlertTypeFilters({ alerts, active: false }),
    alertFlags: toAlertFlagLabels(alerts, alertFlags),
  }
}

```

brightonsbox: Arguably the accumulator is mutating in the reduce though... so is this any better?

server/services/mappers/alertMapper.ts
```diff
@@ -140,3 +147,46 @@ export const toAlertFlagLabels = (alerts: Alert[], alertFlags: AlertFlagLabel[])
...
+            count: acc.inactiveAlertTypesFilter[alert.alertCode.alertTypeCode]
+              ? acc.inactiveAlertTypesFilter[alert.alertCode.alertTypeCode].count + 1
+              : 1,
+            checked: false,
+          },
+        }
+      }
+      return acc
+    },
+    { activeAlertTypesFilter: {}, inactiveAlertTypesFilter: {} },
```
brightonsbox, line 181: Given these two blocks of code (one for active and one for inactive) are basically the same, would it be more readable to convert this to:

```
export const toAlertTypesFilter = ({ alerts, active }: { alerts: Alert[]; active: boolean }) => {
  return alerts
    .filter(alert => alert.isActive === active)
    .reduce((acc, alert) => {
      return {
        ...acc,
        [alert.alertCode.alertTypeCode]: {
          code: alert.alertCode.alertTypeCode,
          description: alert.alertCode.alertTypeDescription,
          count: (acc[alert.alertCode.alertTypeCode] ?? 0) + 1,
          checked: false,
        },
      }
    }, {})
}

export const toAlertSummaryData = (alerts: Alert[], alertFlags: AlertFlagLabel[]): AlertSummaryData => {
  return {
    activeAlertCount: alerts.filter(alert => alert.isActive).length,
    inactiveAlertCount: alerts.filter(alert => !alert.isActive).length,
    activeAlertTypesFilter: toAlertTypesFilter({ alerts, active: true }),
    inactiveAlertTypesFilter: toAlertTypesFilter({ alerts, active: false }),
    alertFlags: toAlertFlagLabels(alerts, alertFlags),
  }
}
```

brightonsbox: (So call the method twice, to get the active and inactive)

brightonsbox (approved):
LGTM

PR #644, opened 2024-06-13
CDPS-761: Past care needs

davidmiddletonmoj:


server/routes/index.ts
```diff
@@ -345,6 +334,16 @@ export default function routes(services: Services): Router {
...
   )
 
+  get(
+    '/prisoner/:prisonerNumber/past-care-needs',
+    auditPageAccessAttempt({ services, page: Page.ReligionBeliefHistory }),
+    getPrisonerData(services, { minimal: true }),
+    checkPrisonerInCaseload(),
+    async (req, res, next) => {
+      return careNeedsController.displayPastCareNeeds(req, res)
+    },
```
whitfield-mj, line 344: I think we've been here before and there was some reason, but this looks the same as just
```suggestion
     careNeedsController.displayPastCareNeeds
    
```

whitfield-mj (approved):

server/routes/index.ts
davidmiddletonmoj: Yeah, can't recall what the reason was it was a while ago, but tried it the suggested way and you get errors.
Might investigate again in a bit, but merging this as-is

PR #653, opened 2024-06-14
CDPS-787: Handle past care needs link when there are no current care needs

davidmiddletonmoj:


server/views/partials/personalPage/careNeeds.njk
```diff
@@ -37,11 +37,14 @@
             {{ summaryListRow(need.type, careNeed(need), "care-need") }}
           {% endfor %}
         </dl>
-        {% if hasPastCareNeeds %}
+      {% endif %}
+      {% if not careNeeds.length and not hasPastCareNeeds %}
+        <p data-qa="no-care-needs-message">No care needs have been entered.</p>
+      {% elseif not careNeeds.length and hasPastCareNeeds %}
```
whitfield-mj, line 43: Really frustrates me that nunjucks has its own language. Why cant they just use JS so I don't have to wonder if something is going to work?!

PR #657, opened 2024-06-19
CDPS-673: Handle non-association API errors

davidmiddletonmoj:
Fix issue with error banner clashing with feedback banner


server/controllers/overviewController.ts
```diff
@@ -114,7 +114,7 @@ export default class OverviewController {
         res.locals.apiErrorCallback,
       ),
       this.offencesService.getOffencesOverview(clientToken, bookingId, prisonerNumber),
-      this.offenderService.getPrisonerNonAssociationOverview(clientToken, prisonerNumber),
+      this.offenderService.getPrisonerNonAssociationOverview(clientToken, prisonerNumber, res.locals.apiErrorCallback),
```
brightonsbox, line 117: I notice that you've not gone down the Result path with this one, why was that? Whichever way we do it I'd rather it be consistent.

server/services/interfaces/offenderService/NonAssociationSummary.ts
```diff
@@ -2,4 +2,5 @@ export default interface NonAssociationSummary {
   prisonName: string
   prisonCount: number
   otherPrisonsCount: number
+  apiError?: boolean
```
brightonsbox, line 5: You can now get rid of this I think?

brightonsbox (approved):
LGTM

brightonsbox (approved):
LGTM

PR #658, opened 2024-06-20
add permissions service for overview router. add guard to restrict ac…

whitfield-mj:
…cess based on service


server/controllers/overviewController.test.ts
```diff
@@ -151,16 +153,20 @@ describe('overviewController', () => {
       )
     })
 
-    it('should not call moneyService.getMoneySummary if prisoner is not in caseload', async () => {
+    it('should not call moneyService.getMoneySummary if user doesnt have money.view permission', async () => {
       const resNotInCaseload = {
```
brightonsbox, line 157: Was in two minds whether we could just use `res` instead of explicitly having the `resNotInCaseload`... since it's now relying on the permissions, it's not necessary is it and bloats the test... meh...

server/controllers/overviewController.ts
```diff
@@ -90,21 +84,18 @@ export default class OverviewController {
...
       this.offencesService.getActiveCourtCasesCount(clientToken, bookingId),
       showCourtCaseSummary ? this.offencesService.getLatestReleaseCalculation(clientToken, prisonerNumber) : null,
-      prisonerInCaseLoad ? this.moneyService.getAccountBalances(clientToken, bookingId) : null,
-      prisonerInCaseLoad || isPomOrReceptionUser
+      permissions.money?.view ? this.moneyService.getAccountBalances(clientToken, bookingId) : null,
+      permissions.adjudications?.view
         ? this.adjudicationsService.getAdjudicationsOverview(clientToken, bookingId)
         : null,
-      prisonerInCaseLoad ? this.visitsService.getVisitsOverview(clientToken, bookingId, prisonerNumber) : null,
+      permissions.visits?.view ? this.visitsService.getVisitsOverview(clientToken, bookingId, prisonerNumber) : null,
```
brightonsbox, line 96: This has changed from just requiring `prisonerInCaseLoad` to `isPrisonerInCaseLoad || userHasRoles([Role.PomUser, Role.ReceptionUser], userRoles)` 

https://github.com/ministryofjustice/hmpps-prisoner-profile/pull/658/files#diff-436cd1a03284961f6db546c6d570d7f2d99d017aacdcf407b84fb925c79cfa47R60

Is this right?

server/services/permissionsService.test.ts
```diff
@@ -0,0 +1,785 @@
...
+        const permissions = await service.getOverviewPermissions(user, PrisonerMockDataA)
+
+        expect(permissions.calculateReleaseDates).toEqual({
+          view: false,
+          edit: true,
+        })
+      })
+    })
+
+    describe('caseNotes', () => {
```
brightonsbox, line 310: I think I perhaps preferred the original parameterised test style for this, think I find it easier to read: 

https://github.com/ministryofjustice/hmpps-prisoner-profile/pull/658/files#diff-f2abab7599c1a00864d045e892ea14706179469fe03671a1ef4e8545fcc891faL39-L50

But that's just personal preference, so happy to let it slide, it looks like all the cases are covered

server/services/permissionsService.test.ts
```diff
@@ -0,0 +1,785 @@
...
+
+        const permissions = await service.getOverviewPermissions(user, PrisonerMockDataA)
+
+        expect(permissions.caseNotes).toEqual({
+          view: false,
+          edit: false,
+        })
+      })
+
+      it('should return edit true if user has GlobalSearch and PomUser roles', async () => {
```
brightonsbox, line 335: Could add `... regardless of case load` in the description perhaps?

server/services/permissionsService.test.ts
```diff
@@ -0,0 +1,785 @@
...
+        const user = {
+          ...prisonUserMock,
+          userRoles: [Role.GlobalSearch, Role.PomUser],
+          caseLoads: [{ caseLoadId: 'LEI', currentlyActive: true } as CaseLoad],
```
brightonsbox, line 339: Or I wonder if the intent would be clearer if it was `caseLoadId: 'SOMETHING_ELSE'` since at first glance we can't see that it's actually different from the prisoner mock data 🤷

server/services/permissionsService.ts
```diff
@@ -0,0 +1,166 @@
...
+          ],
+          userRoles,
+        ),
+      },
+      calculateReleaseDates: {
+        view: false,
+        edit: userHasRoles([Role.ReleaseDatesCalculator], user.userRoles),
+      },
+      caseNotes: {
+        view: false,
```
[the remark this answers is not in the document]
brightonsbox: (same with the other hardcoded `view: false` instances)

server/services/permissionsService.test.ts
```diff
@@ -0,0 +1,785 @@
...
+
+        const permissions = await service.getOverviewPermissions(user, PrisonerMockDataA)
+
+        expect(permissions.caseNotes).toEqual({
+          view: false,
+          edit: false,
+        })
+      })
+
+      it('should return edit true if prisoner is out and user has inactive bookings role', async () => {
```
brightonsbox, line 365: I think we're missing a check of `should return edit false if prisoner is out` (`and user does not have inactive bookings role`)

brightonsbox: Same again for transfer?

I know they're effectively just like any other case load, so if you don't have them you haven't got access, but it might be worth explicitly calling out in the tests?

server/services/permissionsService.ts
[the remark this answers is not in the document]
whitfield-mj: I had assumed that they would all require a view at some point further into this process and was trying to avoid a type where everything is optional as it is pointless. Maybe they won't though so can take out for now

server/controllers/overviewController.test.ts
whitfield-mj: Yeah probably right. The reason I left it was to give an idea to the reader why the permission would be false... but I guess that's not really the point of a unit test

server/controllers/overviewController.ts
whitfield-mj: No, looks like an oversight

server/services/permissionsService.test.ts
whitfield-mj: ok

server/services/permissionsService.test.ts
whitfield-mj: If the prisoner is out and the user doesn't have that role they won't get any permissions, it will bail out at the access section  tested [here](https://github.com/ministryofjustice/hmpps-prisoner-profile/pull/658/files#diff-95688007aec6993ad05ed72087f054153d07188c542f7d7af6f5557d1a305902R80)

server/services/permissionsService.test.ts
```diff
@@ -0,0 +1,785 @@
...
+
+        const permissions = await service.getOverviewPermissions(user, prisoner)
+
+        expect(permissions.useOfForce).toEqual({
+          view: false,
+          edit: true,
+        })
+      })
+
+      it('should return edit true if prisoner is out and user has inactive bookings role', async () => {
```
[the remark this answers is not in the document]
whitfield-mj: Same as above

server/services/permissionsService.test.ts
```diff
@@ -0,0 +1,785 @@
...
+    describe('soc', () => {
+      it.each([Role.SocCommunity, Role.SocCustody])('should return view true if user has %s role', async role => {
+        const user = { ...prisonUserMock, userRoles: [role] }
+
+        const permissions = await service.getOverviewPermissions(user, PrisonerMockDataA)
+
+        expect(permissions.soc.view).toEqual(true)
+      })
+
+      // Odd rule, one to look into
```
brightonsbox, line 657: TODO raise with Joe Cartwright and team?

server/controllers/utils/overviewController/buildOverviewActions.test.ts
```diff
@@ -89,193 +53,197 @@ describe('buildOverviewActions', () => {
...
+        const resp = buildOverviewActions(
+          PrisonerMockDataA,
+          pathfinderNominal,
+          socNominalToUse,
+          prisonUserMock,
+          config,
+          undefined,
+          perms,
+        )
+        expect(
```
brightonsbox, line 182: Since there are so many of these, it might be nice to have a little helper function:

`expectActionVisible(url, visible)`
?

server/controllers/utils/overviewController/buildOverviewActions.ts
```diff
@@ -61,11 +59,7 @@ export default (
     })
   }
 
-  if (
```
brightonsbox, line 64: For JB benefit: I've got to here in the review 😅

server/controllers/utils/overviewController/buildOverviewActions.test.ts
whitfield-mj: Jon now you want me to fix everyone else's work!

server/services/permissionsService.test.ts
```diff
@@ -0,0 +1,785 @@
...
+
+          const permissions = await service.getOverviewPermissions(user, prisoner)
+
+          expect(permissions.accessCode).toEqual(HmppsStatusCode.OK)
+        })
+      })
+    })
+
+    describe('courtCases', () => {
+      it('should return view true if user has ReleaseDatesCalculator role', async () => {
```
brightonsbox, line 132: Do we want to test that it will return false otherwise? I guess I could go and just change the code to:

```
courtCases: {
  view: true,
  edit: true,
}
```

and these tests would still pass

server/services/permissionsService.test.ts
```diff
@@ -0,0 +1,785 @@
...
+      it('should return view true if user has ReceptionUser role', async () => {
+        const user = { ...prisonUserMock, userRoles: [Role.GlobalSearch, Role.ReceptionUser] }
+        const prisoner = { ...PrisonerMockDataA, prisonId: 'LEI' }
+
+        const permissions = await service.getOverviewPermissions(user, prisoner)
+
+        expect(permissions.adjudications).toEqual({
+          view: true,
+        })
+      })
```
brightonsbox, line 205: Test the view false case

server/services/permissionsService.test.ts
```diff
@@ -0,0 +1,785 @@
...
+      it('should return view true if user has ReceptionUser role', async () => {
+        const user = { ...prisonUserMock, userRoles: [Role.GlobalSearch, Role.ReceptionUser] }
+        const prisoner = { ...PrisonerMockDataA, prisonId: 'LEI' }
+
+        const permissions = await service.getOverviewPermissions(user, prisoner)
+
+        expect(permissions.visits).toEqual({
+          view: true,
+        })
+      })
```
brightonsbox, line 237: Test the view false case

server/services/permissionsService.test.ts
```diff
@@ -0,0 +1,785 @@
...
+      it('should return view true if user has GlobalSearch role', async () => {
+        const user = { ...prisonUserMock, userRoles: [Role.GlobalSearch] }
+        const prisoner = { ...PrisonerMockDataA, prisonId: 'LEI' }
+
+        const permissions = await service.getOverviewPermissions(user, prisoner)
+
+        expect(permissions.incentives).toEqual({
+          view: true,
+        })
+      })
```
brightonsbox, line 258: test view false case

server/services/permissionsService.test.ts
whitfield-mj: Have raised with Joe Cartwright. As it is the same as what is currently there I don't think it is a reason to hold up this particular PR, but when I get a response I will update accordingly

server/services/permissionsService.test.ts
brightonsbox: Agreed, but I shouldn't this permissionsService be tested independently of the permissionsGuard? Just thinking since in theory this could be ported over to a separate service sometime in the far future and it'd be good to have a complete set of tests

server/services/permissionsService.test.ts
```diff
@@ -0,0 +1,785 @@
...
+        const user = { ...prisonUserMock, userRoles: [Role.GlobalSearch, Role.ViewProbationDocuments] }
+        const prisoner = { ...PrisonerMockDataA, prisonId: 'LEI' }
+
+        const permissions = await service.getOverviewPermissions(user, prisoner)
+
+        expect(permissions.probationDocuments).toEqual({
+          view: false,
+        })
+      })
+    })
```
brightonsbox, line 752: Think we're missing a test case `should return view true if user has %s role and prisoner is %s`

with OUT and TRN as the final options. since it looks like if they've got the right role and the prisoner is out or transferring they can still view

server/services/permissionsService.test.ts
whitfield-mj: Ok so that means I need to extract all of this and call separately

server/services/permissionsService.test.ts
```diff
@@ -0,0 +1,785 @@
...
+      it('should return PRISONER_IS_TRANSFERRING if prisonId is TRN', () => {
+        const user = { ...prisonUserMock, userRoles: [Role.InactiveBookings, Role.GlobalSearch] }
+        const prisoner = { ...PrisonerMockDataA, restrictedPatient: true, prisonId: 'TRN' }
+
+        const permissions = service.getMoneyPermissions(user, prisoner)
+
+        expect(permissions.accessCode).toEqual(HmppsStatusCode.PRISONER_IS_TRANSFERRING)
+      })
+    })
+  })
```
brightonsbox, line 784: Test the happy path?

brightonsbox (dismissed):
This is excellent - as discussed approving so that you can follow up on the comments in a separate PR

brightonsbox (approved):
LGTM!!

PR #659, opened 2024-06-20
CDPS-774: Height edit routes + controller

danielburnley:


server/views/pages/edit/singleField.njk
```diff
@@ -0,0 +1,64 @@
+{% extends "../index.njk" %}
```
brightonsbox, line 1: Is this in use anywhere?

brightonsbox (approved):
LGTM

server/views/pages/edit/singleField.njk
danielburnley: Not at the moment - it was used by the height page with the intent to be used by the weight page until I got the designs and needed the switcher, figured I'd keep it in for now at least 😅

brightonsbox (approved):
LGTM

PR #672, opened 2024-07-08
CDPS-852: Edit height

danielburnley:


server/controllers/personal/personalController.ts
```diff
@@ -150,10 +166,35 @@ export default class PersonalController {
...
+                Number.isNaN(feet) ||
+                Number.isNaN(inches) ||
+                (feet >= 1 && feet <= 9 && inches < 0),
+              message: 'Feet must be between 1 and 9. Inches must be between 0 and 11',
+            },
+          ]
+
+          const errors = validations.find(({ isInvalid }) => {
+            return isInvalid()
+          })
```
whitfield-mj, line 192: I may be being dim... not sure I follow this. Both validations hardcoded to contain an isInvalid function, so it will always find the first one. As far as I can see errors === validations[0].isInvalid()

server/controllers/personal/personalController.ts
```diff
@@ -150,10 +166,35 @@ export default class PersonalController {
           const feet = parseInt(feetString, 10)
           const inches = parseInt(inchesString, 10)
 
-          if (Number.isNaN(feet) || feet <= 0 || Number.isNaN(inches) || inches <= 0) {
+          const validations = [
+            {
+              isInvalid: () =>
+                ((feetString === '' || feetString === undefined) &&
+                  (inchesString === '' || inchesString === undefined)) ||
```
whitfield-mj, line 173: of these are strings can we not just do `!feetString || !inchesString` ?

whitfield-mj: Also, I think that if it was undefined then parseInt would have already errored and if it was '' then it would have been NaN so perhaps this isn't doing much

server/views/pages/edit/heightMetric.njk
```diff
@@ -4,17 +4,12 @@
...
-  text: 'Digital Prison Services',
-  href:  "/" | prependBaseUrl
-  },
-  {
-  text: breadcrumbPrisonerName,
-  href: "/prisoner/" + prisonerNumber
-  }
+    { text: 'Digital Prison Services', href:  "/" | prependBaseUrl },
+    { text: breadcrumbPrisonerName, href: "/prisoner/" + prisonerNumber }
   ]
```
davidmiddletonmoj, line 13: "Personal" breadcrumb is missing and there is no Cancel option

Same for the Imperial version of the page

server/controllers/personal/personalController.ts
danielburnley: It calls the function so it's the result, if both are false then `find()` errors will be `undefined`

server/controllers/personal/personalController.ts
danielburnley: I had incorrectly presumed '0' gets coerced into false but i've just double checked and that's not the case so I will update this

server/controllers/personal/personalController.ts
```diff
@@ -93,10 +92,26 @@ export default class PersonalController {
           const { editField } = req.body
           const prisonPerson = await this.personalPageService.getPrisonPerson(clientToken, prisonerNumber, true)
 
-          const height = parseInt(editField, 10)
-          if (Number.isNaN(height) || height <= 0) {
+          const height = editField ? parseInt(editField, 10) : 0
+
+          const validations = [
+            {
+              validation: () => editField === '' || editField === undefined || !Number.isNaN(height),
```
whitfield-mj, line 99: As you've already checked this above, perhaps you could just do `editField === 0 || !Number.isNaN(height)`

server/controllers/personal/personalController.ts
danielburnley: I have realised that there's a nicer way to do this anyway so i'm about to change this anyway aha

whitfield-mj (approved):

PR #673, opened 2024-07-10
CDPS-853: Bring weight page in line with designs

danielburnley:


server/controllers/personal/personalController.test.ts
```diff
@@ -52,17 +55,16 @@ describe('PersonalController', () => {
   describe('height', () => {
     describe('metric', () => {
       describe('edit', () => {
-        const action = async (req: any, res: any) => controller.height().metric.edit(req, res, () => {})
+        const action = async (req: any, response: any) => controller.height().metric.edit(req, response, () => {})
 
-        it('Renders the default edit page with the correct data from the prison person API', async () => {
+        fit('Renders the default edit page with the correct data from the prison person API', async () => {
```
brightonsbox, line 60: fit

server/controllers/personal/personalController.ts
```diff
@@ -239,32 +284,52 @@ export default class PersonalController {
...
-          const { editField } = req.body
+          const { kilograms } = req.body
           const prisonPerson = await this.personalPageService.getPrisonPerson(clientToken, prisonerNumber, true)
 
-          const weight = parseInt(editField, 10)
-          if (Number.isNaN(weight) || weight <= 0) {
-            req.flash('fieldValue', editField)
-            req.flash('errors', [{ text: 'Enter a number greater than 0' }])
+          const weight = parseInt(kilograms, 10)
+          const validatedInput = (): { valid: boolean; errorMessage?: string } => {
```
brightonsbox, line 298: It might be worth moving the validation out of the controller and look to see if we can make use of the `validationMiddleware` that we use for case notes already?

brightonsbox: If the validationMiddleware is adapted a bit, we could get it to add to `req.flash` too

server/controllers/personal/personalController.test.ts
danielburnley: 🤦🏻

brightonsbox (approved):
LGTM

PR #679, opened 2024-07-12
CDPS-853: Update validation middleware for use with edit routes

danielburnley:


server/routes/personalRouter.ts
```diff
@@ -129,6 +173,15 @@ export default function personalRouter(services: Services): Router {
...
+        validators: [weightImperialValidator],
+        onError: (req, res) => {
+          const { stone, pounds } = req.body
+          req.flash('stoneValue', stone)
+          req.flash('poundsValue', pounds)
+          return res.redirect(`/prisoner/${req.params.prisonerNumber}/personal/edit/weight/imperial`)
+        },
+      },
     },
   })
```
whitfield-mj, line 187: I wonder if we need to be so specific with the flash messages. If we store `req.flash('inputValues', JSON.stringify(req.body)` then we don't need a specific onError and could just supply a redirect url (if we are unhappy to use the referrer). Then the GET routes would routinely look for flash.inputValues and parse it and the views just look for the same key as the input name. Would make it even less intrusive

server/routes/personalRouter.ts
```diff
@@ -175,12 +164,7 @@ export default function personalRouter(services: Services): Router {
       method: personalController.weight().imperial.submit,
       validation: {
         validators: [weightImperialValidator],
-        onError: (req, res) => {
-          const { stone, pounds } = req.body
-          req.flash('stoneValue', stone)
-          req.flash('poundsValue', pounds)
-          return res.redirect(`/prisoner/${req.params.prisonerNumber}/personal/edit/weight/imperial`)
-        },
+        redirectBackOnError: true,
```
whitfield-mj, line 167: Is this ever not true?

whitfield-mj (approved):

PR #684, opened 2024-07-23
CDPS-851: Edit shoe size

danielburnley:


server/controllers/personal/personalController.ts
```diff
@@ -343,6 +343,85 @@ export default class PersonalController {
...
+          prisonId: prisonerData.prisonId,
+          correlationId: req.id,
+          page: auditPage,
+        })
+
+        res.render('pages/edit/singleField', {
+          pageTitle: `${pageTitle} - Prisoner personal details`,
+          prisonerNumber,
+          breadcrumbPrisonerName: prisonerBannerName,
+          errors: hasLength(errors) ? errors : [],
```
davidmiddletonmoj, line 378: Can just make this return `errors` directly as `req.flash('errors')` will return `[]` if it can't find `errors` in the flash

server/views/pages/edit/singleField.njk
davidmiddletonmoj, line 1: Maybe rename this `textField.njk` just to make it obvious it's only for a text field input.
(I will rename my `radios.njk` to `radioField.njk` in my PR to to fit the theme)

davidmiddletonmoj (approved):

PR #690, opened 2024-07-29
CDPS-741: Disallow global search on location pages

danielburnley:


server/services/utils/permissions/access/getOverviewAccessStatusCode.ts
```diff
@@ -25,5 +29,9 @@ export default function getOverviewAccessStatusCode(user: HmppsUser, prisoner: P
     return globalSearchUser || inactiveBookingsUser ? HmppsStatusCode.OK : HmppsStatusCode.PRISONER_IS_TRANSFERRING
   }
 
+  if (globalSearchUser && !options.allowGlobal) {
+    return HmppsStatusCode.GLOBAL_USER_NOT_PERMITTED
```
davidmiddletonmoj, line 33: What about users that have the right caseload and just happen to also have global search?

Maybe should change to `if (globalSearchUser && !options.allowGlobal && !inUsersCaseload) {`

server/services/permissionsService.ts
```diff
@@ -123,9 +124,12 @@ export default class PermissionsService {
   }
 
   public getLocationPermissions(user: HmppsUser, prisoner: Prisoner): Permissions {
-    const accessCode = getOverviewAccessStatusCode(user, prisoner)
+    const accessCode = getOverviewAccessStatusCode(user, prisoner, { allowGlobal: false })
     if (accessCode !== HmppsStatusCode.OK) return { accessCode }
 
+    // Global is not allowed for locations
```
davidmiddletonmoj, line 130: I don't think this is needed if we change `getOverviewAccessStatusCode` as per other comment?

davidmiddletonmoj (approved):

PR #694, opened 2024-08-01
CDPS-882: Edit smoker/vaper page

danielburnley:


server/controllers/personal/personalController.ts
```diff
@@ -423,6 +424,109 @@ export default class PersonalController {
...
+        const fieldValue =
+          requestBodyFlash?.radioField ||
+          getProfileInformationValue(ProfileInformationType.SmokerOrVaper, inmateDetail.profileInformation)
+
+        // Placeholder for now
+        const options: SelectOption[] = [
+          {
+            text: 'Yes',
+            value: 'Yes',
+            selected: fieldValue === 'Yes',
```
davidmiddletonmoj, line 484: These are radios, so needs to be `checked` to set the radio to "on"
Realised that I've also been using `SelectOption[]` for radios as the govuk design system has the `items` property for selects and radios as the same structure, so generally it just works.
Probably would be good to define a new interface for `RadioOption` (or something like that) so we can put `checked` on there rather than on `SelectOption` which already has `selected`

davidmiddletonmoj (dismissed):

davidmiddletonmoj (approved):

PR #707, opened 2024-08-20
BAVL-256: Use new Book A Video Link API

smcveigh941:
Adds a feature toggle for Book a Video Link functionality.

When the toggle is `true`:
- Video link bookings are created via the Book A Video Link API rather than Whereabouts API.
- Locations offered for the appointment are video link enabled locations only
- "Other" is removed as an option in the list of courts
- An additional field is added to collect the court hearing type (reference data from BAVL API)

### DIFF
--- a/server/controllers/appointmentController.ts
+++ b/server/controllers/appointmentController.ts
@@ -35,6 +35,11 @@ import { ApiAction, AuditService, Page, PostAction, SubjectType } from '../servi
 import config from '../config'
 import logger from '../../logger'
 import { PrisonUser } from '../interfaces/HmppsUser'
+import CreateVideoBookingRequest from '../data/interfaces/bookAVideoLinkApi/CreateVideoBookingRequest'
+import CourtLocation from '../data/interfaces/whereaboutsApi/CourtLocation'
+import Court from '../data/interfaces/bookAVideoLinkApi/Court'
+import VideoLinkLocation from '../data/interfaces/bookAVideoLinkApi/Location'
+import Location from '../data/interfaces/prisonApi/Location'
 
 const PRE_POST_APPOINTMENT_DURATION_MINS = 15
 const { confirmBookingPrisonTemplateId, emails } = config.notifications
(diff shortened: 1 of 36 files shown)


matt-crowson-moj (approved):
LGTM

server/controllers/appointmentController.ts
```diff
@@ -145,7 +159,7 @@ export default class AppointmentController {
         const appointmentsToCreate: AppointmentDefaults = {
           bookingId,
           appointmentType,
-          locationId: Number(location),
+          locationId: appointmentType === 'VLB' && config.featureToggles.bookAVideoLinkEnabled ? vlbLocation : location,
           startTime,
```
timharrison-moj, line 163: Does this this need to be as it was before - with `Number(location)`

server/controllers/appointmentController.ts
```diff
@@ -296,17 +310,27 @@ export default class AppointmentController {
         clientToken,
         user.activeCaseLoadId,
       )
-      courts.push({ id: 'other', name: 'Other' })
+
+      if (!config.featureToggles.bookAVideoLinkEnabled) {
+        ;(courts as CourtLocation[]).push({ id: 'other', name: 'Other' })
+      }
```
timharrison-moj, line 316: semi-colon doesn't look right here?

server/data/interfaces/whereaboutsApi/Appointment.ts
```diff
@@ -33,7 +33,7 @@ export interface AppointmentForm {
 export interface AppointmentDefaults {
   bookingId: number
   appointmentType: string
-  locationId: number
+  locationId: string
   startTime: string
```
timharrison-moj, line 37: Related to the previous question - altered from number to string.

timharrison-moj (approved):
LGTM

server/data/interfaces/whereaboutsApi/Appointment.ts
```diff
@@ -49,11 +49,12 @@ export interface PrePostAppointmentDetails {
...
   formValues?: {
     preAppointment?: string
-    preAppointmentLocation?: number
+    preAppointmentLocation?: string
     postAppointment?: string
-    postAppointmentLocation?: number
+    postAppointmentLocation?: string
     court?: string
     otherCourt?: string
+    hearingType?: string
```
timharrison-moj, line 57: Question really - when BVLS switch is true, we should never let the user provide the `otherCourt` value - they must always select from the list. Do that hold up?

timharrison-moj (dismissed):
LGTM

davidatmoj (approved):

PR #730, opened 2024-09-20
CDPS-949: add distinguishing marks to personal page when prison perso…

whitfield-mj:
…n api is enabled


server/data/interfaces/prisonPersonApi/prisonPersonApiClient.ts
```diff
@@ -102,6 +102,45 @@ export interface PrisonPersonHealthUpdate {
...
+  }
+  partOrientation?: ReferenceDataCodeSimple & {
+    id: 'PART_ORIENT_CENTR' | 'PART_ORIENT_FACE' | 'PART_ORIENT_LOW' | 'PART_ORIENT_UPP'
+  }
+  comment?: string
+  photographUuids?: string[]
+  createdBy: string
+  createdAt: string
+}
+
```
whitfield-mj, line 143: Some part of me wonders if the FE should know about reference data...but have to match on something

server/views/partials/personalPage/distinguishingMarks.njk
```diff
@@ -0,0 +1,66 @@
...
+        {% endfor %}
+    {%  else %}
+        Not entered
+    {% endif %}
+{% endmacro %}
+
+<h3 class="govuk-heading-m">Distinguishing marks</h3>
+<div class='personal-distinguishing-marks'>
+    <div class='personal-distinguishing-marks__image-container govuk-!-padding-left-3 govuk-!-padding-right-8'>
+        <img src='/assets/images/distinguishingMarks/base-person.svg' class='dm-base-image' width='300' height='572' alt=""/>
```
whitfield-mj, line 41: Original plan for accessibility was to add an aria-describedby to a div and build up some text based on what is highlighted. However, as this is pretty much decorative and all the information is contained in the actual text, I think just hiding it is probably better. Open to discussion

brightonsbox (approved):
LGTM!

PR #737, opened 2024-10-16
CDPS-996: Integrate with prison person API for dietary pages

danielburnley:


server/utils/referenceDataCodeUtils.ts
```diff
@@ -0,0 +1,11 @@
+export const formatMedicalDietaryRequirements = (requirements: { id: string; description: string }[]): string[] => {
```
danielburnley, line 1: I've left this in for now, though the `FREE_FROM` label is likely to disappear in the future it's there at the moment - but this can be removed/made more generic once we've got the dietary requirements confirmed

brightonsbox (approved):
LGTM

PR #782, opened 2024-12-05
MAP-1873 Swap to using locations-inside-prison-api for internal locations

GurnankCheema:
The objective behind this code change is to replace calls that get location(s) from prison-api with calls to locations-inside-prison-api. 

The specific api endpoints changed are called by these functions in locationsInsidePrisonApiClient.ts. They have the same name as previously in prisonApiClient.
- getLocation
- getLocationsForAppointments

The key pages affected are 
- prisoner/{prisonNumber}/add-appointment
- prisoner/{prisonNumber}/appointment-confirmation

The nomis-sync-prisoner-mapping-api is used to get the corresponding location id between the prison-api and locations-inside-prison-api

Where used, dpsLocationId refers to the new uuid type id returned in the new locationsInsidePrisonApi object and nomisLocationId refers to the (current) integer type id  returned in the prisonApi object

Data passed to the backend for persisting will still use the Nomis version of location id. The nomis-sync-prisoner-mapping-api is used to find the nomisLocationId from the dpsLocationId and vice-versa.

[locations api swagger](https://locations-inside-prison-api-dev.hmpps.service.justice.gov.uk/swagger-ui/index.html?configUrl=/v3/api-docs)

[sync api swagger](https://nomis-sync-prisoner-mapping-dev.hmpps.service.justice.gov.uk/webjars/swagger-ui/index.html)

### DIFF
--- a/assets/js/prePostAppointment.js
+++ b/assets/js/prePostAppointment.js
@@ -4,10 +4,9 @@ document.addEventListener('DOMContentLoaded', () => {
   const postAppointmentEventsContainer = document.getElementById('postAppointmentEventsContainer')
   const postAppointmentLocation = document.getElementById('postAppointmentLocation')
   const courtSelect = document.getElementById('court')
-  const otherCourtContainer = document.getElementById('otherCourtContainer')
 
   async function getPrePostEventsForLocation(input, container) {
-    const locationId = Number(input.value)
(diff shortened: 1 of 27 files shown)


brightonsbox (approved):
LGTM

PR #788, opened 2024-12-11
CDPS-1081: Updating NodeJS to 22.12

brightonsbox:


patches/@apollo+client+3.11.10.patch
davidatmoj, line 1: in `package.json` we have an update to 3.12.2, so not sure this is valid?
`"@apollo/client": "^3.12.2"`

server/utils/dateHelpers.test.ts
```diff
@@ -104,7 +104,7 @@ describe('format date', () => {
     ['[default]', '2023-01-20', undefined, '20 January 2023'],
     ['long', '2023-01-20', 'long', '20 January 2023'],
     ['short', '2023-01-20', 'short', '20/01/2023'],
-    ['full', '2023-01-20', 'full', 'Friday 20 January 2023'],
+    ['full', '2023-01-20', 'full', 'Friday, 20 January 2023'],
```
davidatmoj, line 107: This means the formatting of full dates will not be correct in the app. They should not have a comma (see screenshot)
This behaviour was present in a previous version of node, where we had to remove the comma during formatting, but changed to not display the comma apparently in v22. (I wished they'd make their minds up!)
I can't seem to find the commit where we removed the formatting change.

image

patches/@apollo+client+3.11.10.patch
brightonsbox: Hmm, it still fixed the import issue, not sure how though! I've updated and pinned the version of the client for now so they don't get out of step

package.json
```diff
@@ -29,7 +29,8 @@
     "int-test": "cypress run --config video=false",
     "int-test-ui": "cypress open --e2e --browser chrome",
     "clean": "rm -rf dist build node_modules stylesheets",
-    "rebuild": "npm run clean && npm i && npm run build"
+    "rebuild": "npm run clean && npm i && npm run build",
+    "postinstall": "npx patch-package"
   },
   "engines": {
     "node": "^20",
```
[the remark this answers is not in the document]
brightonsbox: Good spot, thank you!

server/utils/dateHelpers.test.ts
brightonsbox: Thanks David, that should be resolved now

server/utils/dateHelpers.ts
```diff
@@ -104,14 +106,16 @@ export const isRealDate = (date: string): boolean => {
 export const formatDate = (isoDate: string, style: 'short' | 'full' | 'long' | 'medium' = 'long'): string => {
   if (!isoDate) return ''
 
-  return new Date(isoDate).toLocaleDateString('en-gb', { dateStyle: style })
+  return new Date(isoDate).toLocaleDateString('en-gb', { dateStyle: style })?.replaceAll(',', '')
```
brightonsbox, line 109: Removing unwanted commas

PR #789, opened 2024-12-12
BAVL-462: Remove BAVL feature toggle, remove gov notify, and tidy up locations selection

smcveigh941:
- Removes the `bookAVideoLinkEnabled` feature toggle
- Removes any legacy code related to VLB appointments now that the toggle is permanently on
- Removes integration with GOV Notify
- "Video link locations" are now the same list of locations as regular appointment locations, so removed any code which separated them out, including some client side JS
- Integrated with Locations API and the Nomis Mapping client in order to map Nomis location IDs to Location API business keys in order to post locations to BVLS API


davidatmoj (approved):
LGTM

PR #790, opened 2024-12-13
CDPS-997 & CDPS-998: Change body part & location

davidatmoj:


integration_tests/tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es2018",
+    "target": "es2022",
```
brightonsbox, line 3: Not really important, but I think we can do the whole:

```
"extends": "@tsconfig/node22/tsconfig.json"
```
thing here too

brightonsbox (approved):
LGTM

PR #794, opened 2024-12-17
BAVL-421: Enable appointments of type VLB to be edited

smcveigh941:
- Adds the ability to amend a video link booking
- This does not cover the edit of other types of appointment since a prison API endpoint does not exist for that, but the code has been written to make that happen easily if required later
- The amend journey is on a different route with the appointment ID in the URL, but it uses the same route handler (in "edit mode" when an appointment ID is provided)


brightonsbox (approved):
LGTM

PR #801, opened 2025-01-06
CDPS-1088 Edit nationality

scottrowley:


server/data/personIntegrationApiClient.ts
```diff
@@ -14,10 +14,18 @@ export default class PersonIntegrationApiRestClient implements PersonIntegration
...
 
   updateBirthPlace(prisonerNumber: string, birthPlace: string): Promise<void> {
+    return this.updateCorePersonRecord(prisonerNumber, 'BIRTHPLACE', birthPlace)
+  }
+
+  updateNationality(prisonerNumber: string, nationality: string): Promise<void> {
+    return this.updateCorePersonRecord(prisonerNumber, 'NATIONALITY', nationality)
+  }
+
+  private updateCorePersonRecord(prisonerNumber: string, fieldName: string, value: string): Promise<void> {
```
brightonsbox, line 24: Nice spot that we can reuse this. We can apply this to `updateCountryOfBirth` too

server/services/personalPageService.ts
```diff
@@ -216,6 +216,11 @@ export default class PersonalPageService {
       }
     }
 
+    let nationality = 'Not entered'
+    if (inmateDetail?.profileInformation?.some(entry => entry.type === 'NAT')) {
+      nationality = inmateDetail.profileInformation.filter(entry => entry.type === 'NAT')[0].resultValue
+    }
```
brightonsbox, line 222: Maybe this could be:

```
const nationality = inmateDetail?.profileInformation
    ?.find(entry => entry.type === 'NAT')
    ?.resultValue 
    || 'Not entered'
```

server/services/personalPageService.ts
```diff
@@ -552,4 +557,17 @@ export default class PersonalPageService {
 
     return response
   }
+
+  async updateNationality(clientToken: string, user: PrisonUser, prisonerNumber: string, nationality: string) {
+    const personIntegrationApiClient = this.personIntegrationApiClientBuilder(clientToken)
+    const response = personIntegrationApiClient.updateNationality(prisonerNumber, nationality)
```
brightonsbox, line 563: I can see that we've missed it in a few other places too but we could do to do an `await` on the api call so that the metrics only get registered if the update is successful.

brightonsbox (approved):
LGTM

PR #802, opened 2025-01-06
BAVL-547 edit appointment h1

timharrison-moj:
Small change to show an alternative page H1 during editing of a video link booking.
(Only VLB type appointments can be edited via the profile - other types can only be cancelled)


brightonsbox (approved):
LGTM

PR #832, opened 2025-02-05
CDPS-1174: Correcting audit action types

brightonsbox:


server/controllers/personal/fieldData.ts
```diff
@@ -2,19 +2,20 @@ import {
...
 import { formatHeight, formatWeight } from '../../utils/utils'
 
 export interface FieldData {
+  url: string
   fieldName: string
   pageTitle: string
-  auditPage: Page
-  url: string
+  auditEditPageLoad: Page
+  auditEditPostAction: PostAction
```
brightonsbox, line 13: Keeping the audit post action alongside this field data

server/controllers/personal/personalController.test.ts
```diff
@@ -291,7 +291,7 @@ describe('PersonalController', () => {
             user: prisonUserMock,
             prisonerNumber: 'ABC123',
             correlationId: request.id,
-            action: PostAction.EditPhysicalCharacteristics,
+            action: PostAction.EditHeight,
```
brightonsbox, line 294: Since we're no longer editing physical characteristics in the Prison Person API, and we edit each field individually I thought it made sense to actually have a separate action for each edit

brightonsbox: OR we could make these actions all more generic, maybe EDIT_PROFILE perhaps? Because we record the fieldName in the details 🤷

davidatmoj (approved):

PR #843, opened 2025-02-12
CDPS-1114: Change how main photo is flagged

danielburnley:


server/services/photoService.ts
```diff
@@ -11,14 +11,14 @@ export default class PhotoService {
     return prisonApiClient.getImageDetail(prisonerPhotoId)
   }
 
-  async getAllFacialPhotos(prisonerNumber: string, clientToken: string) {
+  async getAllFacialPhotos(prisonerNumber: string, prisonerPhotoId: number, clientToken: string) {
```
brightonsbox, line 14: Maybe as a follow up it might be nice to call this `mainPrisonerPhotoId` to make it clear what the intention is behind passing the id in?

brightonsbox (approved):
LGTM

server/services/photoService.ts
danielburnley: Yeah - that makes sense. I'll do that in my next follow up

PR #848, opened 2025-02-17
CDPS-1224: Enforcing access only to prison (NOMIS) users

brightonsbox:


server/app.ts
```diff
@@ -43,7 +43,7 @@ export default function createApp(services: Services): express.Application {
   nunjucksSetup(app, path)
   setUpEnvironmentName(app)
   app.use(setUpAuthentication())
-  app.use(authorisationMiddleware(['ROLE_PRISON', 'ROLE_GLOBAL_SEARCH']))
+  app.use(authorisationMiddleware(['ROLE_PRISON']))
```
brightonsbox, line 46: The only non-test change...

PR #853, opened 2025-02-21
CDPS-1158 & CDPS-1159: Use proxy for physical attributes

davidatmoj:


server/services/personalPageService.ts
```diff
@@ -105,22 +97,35 @@ export default class PersonalPageService {
...
     prisonerNumber: string,
-    physicalAttributes: Partial<PrisonPersonPhysicalAttributesUpdate>,
+    physicalAttributes: CorePersonPhysicalAttributesRequest,
   ) {
-    const apiClient = this.prisonPersonApiClientBuilder(token)
-    const response = await apiClient.updatePhysicalAttributes(prisonerNumber, physicalAttributes)
+    const apiClient = this.personIntegrationApiClientBuilder(token)
+
+    const existingPhysicalAttributes =
+      (await apiClient.getPhysicalAttributes(prisonerNumber)) ??
```
brightonsbox, line 114: By this point we've already retrieved the physical attributes once in the controller so we're ending up making the same call twice? Maybe it's fine, we're not going to have loads of these requests...

brightonsbox (approved):
LGTM

PR #896, opened 2025-04-01
CDPS-1099 Edit sexual orientation

scottrowley:


server/services/auditService.ts
```diff
@@ -9,101 +9,103 @@ export enum Page {
   ActiveAlerts = 'ACTIVE_ALERTS',
   AddAlert = 'ADD_ALERT',
   AddAppointment = 'ADD_APPOINTMENT',
-  EditAppointment = 'EDIT_APPOINTMENT',
   AddCaseNote = 'ADD_CASE_NOTE',
   AddDistinguishingMark = 'ADD_DISTINGUISHING_MARK',
   AddDistinguishingMarkPhoto = 'ADD_DISTINGUISHING_MARK_PHOTO',
+  AddMilitaryServiceInformation = 'ADD_MILITARY_SERVICE_INFORMATION',
```
scottrowley, line 15: Mostly reordering - EditSexualOrientation and Post EditSexualOrientation are the new values

server/views/macros/sexualOrientationText.njk
```diff
@@ -0,0 +1,13 @@
+{%- macro sexualOrientationText(orientation) -%}
```
davidatmoj, line 1: Maybe for another PR, but would be good to consolidate the description mapping into one place (e.g. `descriptionLookup` in the controller), and turn this into a filter using the same lookup rather than a macro

davidatmoj (approved):

PR #903, opened 2025-04-03
BAVL-744: Strings in the body of forms are trimmed before validation takes place

smcveigh941:
A validation error was throwing on production for an invalid email address, where the user had entered trailing spaces at the end of the field. Generally, form fields should be trimmed before any validation takes place.


brightonsbox (approved):
LGTM

PR #911, opened 2025-04-04
CDPS-944: Update other languages

davidatmoj:


assets/js/languages.js
```diff
@@ -1,31 +1,35 @@
-document.querySelectorAll('.js-autocomplete-select').forEach(selectElement => {
-  accessibleAutocomplete.enhanceSelectElement({
-    defaultValue: '',
-    selectElement: selectElement,
-    inputClasses: 'govuk-input--width-30',
-    menuClasses: 'govuk-input--width-30',
+document.addEventListener('DOMContentLoaded', function () {
```
brightonsbox, line 1: We have a generic `autocomplete.js` that was used for country of birth, could this be reused instead? https://github.com/ministryofjustice/hmpps-prisoner-profile/blob/main/assets/js/autocomplete.js

server/services/languagesService.ts
```diff
@@ -52,4 +53,33 @@ export default class LanguagesService {
...
+    user: PrisonUser,
+    prisonerNumber: string,
+    secondaryLanguageRequest: SecondaryLanguageRequest,
+  ) {
+    const personCommunicationNeedsApiClient = this.personCommunicationNeedsApiClientBuilder(clientToken)
+
+    await personCommunicationNeedsApiClient.updateSecondaryLanguage(prisonerNumber, secondaryLanguageRequest)
+
+    this.metricsService.trackPersonCommunicationNeedsUpdate({
+      fieldsUpdated: ['secondary-language'],
```
brightonsbox, line 68: I think we've got a bit of a mix of kebab case and camel case field names for metrics - would be good to go with one and be consistent

assets/js/languages.js
brightonsbox: I just wondered if this would cater for just one since `document.querySelectorAll('.js-autocomplete-select')` would just return an array of size one?

brightonsbox (approved):
LGTM

assets/js/languages.js
davidatmoj: Yeah, there was some issue I came across and I can't remember now, but I did tweak my version a few times, so maybe it would work. I'll have a go.

PR #921, opened 2025-04-10
CDPS-993 & CDPS-994: Update validation for languages

davidatmoj:


davidatmoj:
Dev notes: Based on UCD feedback, I've removed existing languages from the autocomplete list so they cannot be selected. If the user manually types a matching existing language (case insensitive), it still triggers a "duplicate" error. If they type an invalid language it triggers an "invalid" error. Due to these changes, most validation checks are now done client-side, and errors sent to the server for processing.
This saves some extra API calls that were previously made, and in the case JS is not available in the UI, these errors cannot happen anyway as we use a select control.
An additional enhancement is that now the invalid (or duplicate) language is shown in the input field when displaying the errors.

scottrowley (approved):

PR #922, opened 2025-04-10
CDPS-1342: update to use latest dockerFile

malaw-moj:


Dockerfile
```diff
@@ -25,30 +29,21 @@ ENV BUILD_NUMBER=${BUILD_NUMBER}
...
 ARG BUILD_NUMBER
 ARG GIT_REF
 ARG GIT_BRANCH
 
 COPY package*.json ./
 RUN CYPRESS_INSTALL_BINARY=0 npm ci --no-audit
+ENV NODE_ENV='production'
 
 COPY . .
-RUN npm run postinstall
```
brightonsbox, line 44: We need to keep this - this patches the apollo-client dependency

Dockerfile
```diff
@@ -58,12 +53,6 @@ COPY --from=build --chown=appuser:appgroup \
         /app/package-lock.json \
         ./
 
-COPY --from=build --chown=appuser:appgroup \
-        /app/build-info.json ./dist/build-info.json
-
-COPY --from=build --chown=appuser:appgroup \
-        /app/assets ./assets
-
```
brightonsbox, line 66: You'll need to keep these too.  The build info stuff for example has changed in the template, it's now pulled in a different way to how we do it: https://github.com/ministryofjustice/hmpps-template-typescript/blob/main/server/config.ts#L32-L35

malaw-moj: Resolved

brightonsbox (approved):
LGTM

PR #928, opened 2025-04-16
CDPS-1200: Next of kin and emergency contacts on personal page

davidatmoj:


server/views/partials/personalPage/nextOfKin.njk
```diff
@@ -1,55 +1,100 @@
...
+      <div class="govuk-grid-column-full">
+        {% if newArrival72 %}
+          <div class="not-entered-tag not-entered-tag--heading">Not entered</div>
+        {% endif %}
+        <h3 class="govuk-!-font-size-24 govuk-!-margin-top-0 govuk-!-margin-bottom-3">No emergency contact</h3>
+        <p>No contact details have been entered.</p>
+      </div>
+    </div>
+  {% endif %}
+  {% if hasNextOfKin != hasEmergencyContact %}
```
scottrowley, line 64: Should this be `!==`?

Unless I'm misunderstanding the logic (I probably am), this can only happen when `nextOfKinAndEmergencyContacts.length > 0`, so we could move the `<hr>` to the top of the loop and remove this check and the `loop.last` check.

scottrowley (approved):

server/views/partials/personalPage/nextOfKin.njk
davidatmoj: Yeah, that's another typo... should be `!==`

But in terms of the logic - if we have BOTH NoK and EC, we don't show a HR here, so we need this condition to only show it if there is ONE of the "No contact" states displayed.

PR #960, opened 2025-04-30
CDPS-1240: Photo edit styling

danielburnley:


server/views/pages/edit/photo/editPhoto.njk
```diff
@@ -9,18 +9,53 @@
...
+  }
+] %}
+{% set cropHtml %}
+<div class="hmpps-image-cropper-button-text">
+  <img src="/assets/images/imageEditing/crop.svg" role="presentation" alt=""/> Crop
+</div>
+{% endset %}
+{% set previewHtml %}
+<div class="hmpps-image-cropper-button-text">
+  <img src="/assets/images/imageEditing/crop.svg" role="presentation" alt=""/> Preview
```
scottrowley, line 33: Is this intentional until we have a preview.svg?

assets/js/profilePhotoCropper.js
```diff
@@ -76,7 +77,7 @@ function setCropperListener(fileName, fileType) {
   const canvas = document.querySelector('cropper-canvas')
   const selector = document.querySelector('cropper-selection')
   canvas.addEventListener('actionend', actionEvent => {
-    selector.$toCanvas().then(canvas => {
+    selector.$toCanvas({ width: 427, height: 570 }).then(canvas => {
```
scottrowley, line 80: Why do these differ from the max values above?

MeliaBurnley: Ah yeah I'll add a comment for this - this is what the image gets scaled to in the API, without specifying the size here it will automatically do it based on the viewport size of the selection so it gets scaled about weirdly

MeliaBurnley, line 80:
```suggestion
    // The width and height the image gets scaled to in the API, this prevents the image becoming pixellated
    selector.$toCanvas({ width: 427, height: 570 }).then(canvas => {
```

scottrowley (approved):

PR #972, opened 2025-05-08
BAVL-840 this change removes the master VLPM appointment types feature toggle. The code sitting behind this is now safe to use, the old code can go.

matt-crowson-moj:
This change removes the master VLPM feature toggle from the service.

The code that was sitting behind it is now always used going forwards.  The old code is no longer needed so it has been removed.

### DIFF
--- a/server/controllers/appointmentController.ts
+++ b/server/controllers/appointmentController.ts
@@ -40,7 +40,6 @@ import CreateVideoBookingRequest, {
   VideoLinkBooking,
 } from '../data/interfaces/bookAVideoLinkApi/VideoLinkBooking'
 import NomisSyncLocation from '../data/interfaces/nomisSyncPrisonerMappingApi/NomisSyncLocation'
-import { bvlsMasteredVlpmFeatureToggleEnabled } from '../utils/featureToggles'
 
 const PRE_POST_APPOINTMENT_DURATION_MINS = 15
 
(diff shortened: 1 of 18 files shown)


server/controllers/appointmentController.ts
```diff
@@ -424,15 +410,15 @@ export default class AppointmentController {
             ])
 
             return {
-              bookingType: !bvlsMasteredVlpmFeatureToggleEnabled() ? vlb.bookingType : undefined,
+              bookingType: undefined as string,
```
timharrison-moj, line 413: Not sure why we wouldn't put the bookingType here? But agree that this is what the code was doing when the toggle is unset.

timharrison-moj: Unless this is a court booking!

timharrison-moj: Which it is, of course.

server/validators/prePostAppointmentValidator.ts
```diff
@@ -28,27 +27,15 @@ export const PrePostAppointmentValidator: Validator = (body: Record<string, stri
     errors.push({ text: 'Select a room for the post-court hearing briefing', href: '#postAppointmentLocation' })
   }
 
-  if (!bvlsMasteredVlpmFeatureToggleEnabled() && !body.bookingType) {
-    errors.push({ text: 'Select a booking type', href: '#bookingType' })
-  }
-
-  if ((bvlsMasteredVlpmFeatureToggleEnabled() || body.bookingType === 'COURT') && !body.court) {
+  if (!body.court) {
```
timharrison-moj, line 30: I don't see the check for `body.bookingType == 'COURT'` in the above check anymore? It was a switch OR, so maybe always triggered now, hence removed. Probably right.

timharrison-moj (dismissed):
LGTM.
Quite a few brain-teasing boolean logic puzzles in this one!

server/controllers/appointmentController.ts
matt-crowson-moj: Agree, though it does look/read strange.

matt-crowson-moj:
> LGTM. Quite a few brain-teasing boolean logic puzzles in this one!

Yeah I was a little uncertain with these, but local testing/running seemed ok.  We need some regression testing for this one.

assets/js/addAppointment.js
```diff
@@ -73,11 +73,6 @@ document.addEventListener('DOMContentLoaded', () => {
   function showHideProbationFields() {
     const appointmentType = appointmentTypeSelect.value
 
-    // TODO remove this when the feature toggle BVLS_MASTERED_VLPM_FEATURE_TOGGLE_ENABLED is removed
```
matt-crowson-moj, line 76: @timharrison-moj  FYI

timharrison-moj (dismissed):
LGTM

PR #982, opened 2025-05-12
BAVL-818: Resolve an issue with flash after a genuine API error

timharrison-moj:
Tricky problem to find, test and resolve.

After an API error (e.g. preventing a duplicate appointment) it was possible that flash keys could contain multiple sets of data, and this could be confusing to users as it would contain the details of the last failed booking, rather than the current one. 

This ensures that at most one set of data (the current booking) is held in flash between pages.


matt-crowson-moj (approved):
LGTM

PR #983, opened 2025-05-13
CDPS-1430 & CDPS-1429: Update pages based on feedback

MeliaBurnley:
- selection is now centred by default
- selection is now the govuk focus colour
- the user flow is changed to be more clear
- fixed issue with photo scale when rotating and not moving
  - Removed duplicate code as a result of this


brightonsbox (approved):
LGTM

PR #985, opened 2025-05-13
CDPS-1417: Fix rules around displaying 'Add key worker session' link

brightonsbox:


server/controllers/utils/overviewController/buildOverviewActions.ts
```diff
@@ -41,7 +41,7 @@ export default (
     })
   }
 
-  if (permissions.keyWorker?.edit) {
+  if (user.authSource === 'nomis' && user.keyWorkerAtPrisons?.[prisonerData.prisonId]) {
```
mtac50, line 44: very minor point - Do we need to check the auth source? This check is essentially covered by the key worker check

mtac50 (approved):
Minor comment but nothing I'd hold up approval over :)

LGTM

PR #988, opened 2025-05-13
BAVL-847: Dev allow list only

timharrison-moj:
Problem: 
- Some internal staff have MOJ windows laptops & a VPN which provides source IPs in `prisons` IP allowlist group. 
- They currently get a 403 error for any profile links in the DEV environment
- Adding this group to the allow list in DEV will mean these team members can use DEV effectively. 
- This group is already present for preprod and prod, so just extending to DEV too.


brightonsbox (approved):
LGTM

PR #1010, opened 2025-05-23
MIA-355: Add warning text to KW Case Notes journey

JackReeveMoJ:
New warning text shows when selecting Key Worker Activity case note types. Fixed a bug where the warning text did not disappear after selecting another type that should not show warning text the first time

image

image


brightonsbox (approved):
LGTM

PR #1035, opened 2025-06-10
CDPS-1380: Adding confirm address page

brightonsbox:


integration_tests/e2e/personal/edit/address/confirmAddress.cy.ts
```diff
@@ -0,0 +1,111 @@
...
+    countryCode: 'COUNTRY1',
+    postCode: 'A1 2BC',
+    fromDate: '2025-06-10',
+    addressTypes: [],
+  }
+  const cachedAddress = JSON.stringify({ address, route: 'find-uk-address' })
+
+  before(() => {
+    // Seed redis data:
+    cy.exec(`./integration_tests/scripts/redis-cli "SET ${addressKey} '${cachedAddress.replaceAll('"', '\\"')}'"`)
```
MeliaBurnley, line 30: I'm not sure if this will ever be needed elsewhere so probably not necessary, but if it was i'd probably:

- Make this a cypress command so that it's clear what it's doing (maybe something like `cy.redisSet(key, value)`
- Make `cachedAddress.replaceAll(...)` a variable with a descriptive name so that it doesn't require parsing the regex if you don't need to

MeliaBurnley (approved):

PR #1040, opened 2025-06-12
CDPS-1458 Adding precommit hooks 0.0.2

andrewrlee:
This seems to be working now:
```sh
docker build -t aaa:01 --build-arg BUILD_NUMBER=0.0.1 --build-arg GIT_REF=aaaa --build-arg GIT_BRANCH=main 
```
<kbd>
Screenshot 2025-06-12 at 09 51 03
</kbd>


andrewrlee:
circle ci runs `dash` if you run `/bin/sh` which doesn't support pipeline 😩

PR #1057, opened 2025-06-20
CDPS-1286 Edit existing identity numbers

scottrowley:


server/routes/identityNumbersRouter.ts
```diff
@@ -3,49 +3,72 @@ import { getRequest, postRequest } from './routerUtils'
...
-    validationMiddleware([identityNumbersValidator], {
+    validationMiddleware([addIdentityNumbersValidator], {
+      redirectBackOnError: true,
+    }),
+    identityNumbersController.homeOfficeIdNumbers().submit,
+  )
+
+  // Edit individual existing ID numbers
+  get(
+    `/:identifier(${identityNumberRoutes})/:compositeId`,
```
scottrowley, line 61: Routes look like /personal/cro/123-4. This is a composite of the offenderId and sequenceId.

Routing with /cro, /pnc was requested by content, but it doesn't actually provide function, only the composite id is needed.

brightonsbox (approved):
LGTM

PR #1063, opened 2025-06-20
BAVL-904: Feature-switched updates to the prisoner profile create video court appointment journey

timharrison-moj:
This PR adds an optional way to enter the court video link, and an optional guest pin to access it.
Tested locally with the feature both on and off - defaults to false.
This will turn on the feature for DEV, but leave it false for PREPROD and PROD.

With feature on, there are extra choices on the pre-post-appointments form for COURT bookings:
image

And these are reflected in the confirmation form for the booking too:
image

The journey remains unchanged with the feature switch off.


matt-crowson-moj (approved):
LGTM

PR #1068, opened 2025-06-26
CDPS-1475 Reduce contacts API fetch size from 1000 to 50

scottrowley:


server/config.ts
```diff
@@ -363,7 +363,6 @@ export default {
     editProfileEnabled: toBoolean(get('EDIT_PROFILE_ENABLED', 'false')),
     editProfileEnabledPrisons: get('EDIT_PROFILE_ENABLED_PRISONS', []),
 
-    // TODO Remove once we're sure that the personal relationships API is stable in production (enabled w/c 23rd June 2025)
```
scottrowley, line 366: https://dsdmoj.atlassian.net/browse/CDPS-1475 asks if we can maintain the prison API fallback for now

PR #1081, opened 2025-07-03
CDPS-1485: Report when locked warning happens

Ewan-Donovan:
I think this is the right approach, capturing detail here instead of tracking it on each separate proxy api.

I figured it also makes sense to track which page the warning happened on, and which api url returned the locked status. This way we can see if there's a common situation where the locked warning is occurring.

- created a new method on the metrics service
- passed the metrics service into the lock-handling middleware which then calls it
- updated tests


brightonsbox (approved):
LGTM

PR #1083, opened 2025-07-04
CDPS-1491: First pass refactoring to shared edit page template

MeliaBurnley:


server/utils/featureToggles.ts
```diff
@@ -13,8 +13,7 @@ export const militaryHistoryEnabled = () =>
   config.featureToggles.militaryHistoryEnabledFrom &&
   isAfter(Date.now(), config.featureToggles.militaryHistoryEnabledFrom)
 
-export const editReligionEnabled = () =>
-  config.featureToggles.editReligionEnabledFrom && isAfter(Date.now(), config.featureToggles.editReligionEnabledFrom)
+export const editReligionEnabled = () => true
```
brightonsbox, line 16: Can we just rip this out completely?

MeliaBurnley: Ah yeah this was just for local dev because mine was disabled and I was being quick, but yes we can rip both this and the military history code out now, can probably do this as a followup/separate ticket though as this is big enough 😅

Ewan-Donovan (approved):
LGTM - always nice to see a PR with more lines removed than added

PR #1095, opened 2025-07-08
BAVL-902 small content fix for the guest pin radio and video link label.

matt-crowson-moj:
Small context fix for BVLS appointments:

Before:
dps-before

After:
dps-after


timharrison-moj (approved):
LGTM

PR #1147, opened 2025-07-18
Refactor professionalContactsSerivce to use keyworker-api only for key worker name and last session

bill-moj:
…

### DIFF
--- a/server/data/localMockData/keyWorker.ts
+++ b/server/data/localMockData/keyWorker.ts
@@ -1,4 +1,5 @@
 import KeyWorker from '../interfaces/keyWorkerApi/KeyWorker'
+import StaffAllocation from '../interfaces/keyWorkerApi/StaffAllocation'
 
 export const keyWorkerMock: KeyWorker = {
   staffId: 3532453,
@@ -7,6 +8,29 @@ export const keyWorkerMock: KeyWorker = {
   email: '1@1.com',
 }
 
+export const StaffAllocationMock: StaffAllocation = {
+  hasHighComplexityOfNeeds: false,
+  allocations: [
+    {
+      policy: {
+        code: 'KEY_WORKER',
+        description: 'Key worker',
+      },
+      prison: {
+        code: 'CODE',
+        description: 'Description',
+      },
+      staffMember: {
+        staffId: 3532453,
+        firstName: 'Dave',
+        lastName: 'Stevens',
+      },
+    },
+  ],
+  latestRecordedEvents: [],
+}
+
 export default {
   keyWorkerMock,
+  StaffAllocationMock,
 }
--- a/server/services/professionalContactsService.ts
+++ b/server/services/professionalContactsService.ts
@@ -11,7 +11,11 @@ import {
   SortType,
 } from '../utils/utils'
 import Address from '../data/interfaces/prisonApi/Address'
-import StaffContacts, { Contact, YouthStaffContacts } from '../data/interfaces/prisonApi/StaffContacts'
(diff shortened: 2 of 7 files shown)


server/data/localMockData/keyWorker.ts
```diff
@@ -7,6 +8,38 @@ export const keyWorkerMock: KeyWorker = {
   email: '1@1.com',
 }
 
+export const StaffAllocationMock: StaffAllocation = {
```
iterative-it, line 11: Super minor - rest of the mocks in here are named camel case

server/services/professionalContactsService.ts
```diff
@@ -189,12 +189,20 @@ export default class ProfessionalContactsService {
...
     const isYouthPrisoner = youthEstatePrisons.includes(prisonId)
     return isYouthPrisoner
       ? this.getYouthStaffContactsOverview(clientToken, bookingId)
-      : this.getStaffContactsOverview(clientToken, bookingId, prisonerNumber, prisonId, apiErrorCallback)
+      : this.getStaffContactsOverview(
+          clientToken,
+          bookingId,
+          prisonerNumber,
+          prisonId,
+          isNewKeyWorkerServiceEnabled,
```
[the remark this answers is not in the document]
bill-moj: the plan is to get rid of this boolean and divert all calls to the new endpoint after we see that things are smooth after the new key worker service rollout? i don't expect this to stay after personal officer allocation service is released.

server/services/professionalContactsService.ts
```diff
@@ -243,12 +251,58 @@ export default class ProfessionalContactsService {
...
+        Result.wrap(allocationManagerApiClient.getPomByOffenderNo(prisonerNumber), apiErrorCallback),
+        Result.wrap(keyWorkerClient.getCurrentAllocations(prisonerNumber), apiErrorCallback),
+        prisonApi.getBookingContacts(bookingId),
+      ])
+
+      return {
+        keyWorker: currentAllocatedStaff
+          .map(({ allocations, latestRecordedEvents, hasHighComplexityOfNeeds }) => {
+            const allocatedKeyWorker = allocations.find(itm => itm.policy.code === 'KEY_WORKER')?.staffMember
+            const lastKeyWorkerSession = latestRecordedEvents.find(itm => itm.type === 'SESSION')
```
iterative-it, line 275: I don't think `latestRecordedEvents` can be null but might be worth adding `?` anyway for extra prisoner profile safety. Also theoretically possible there could be more than one `SESSION`. Maybe add a first or default equivalent?

server/services/professionalContactsService.ts
```diff
@@ -243,12 +251,58 @@ export default class ProfessionalContactsService {
...
+        prisonOffenderManager: allocationManager
+          .map(pom => formatPomName(pom?.primary_pom?.name))
+          .toPromiseSettledResult(),
+        coworkingPrisonOffenderManager: allocationManager
+          .map(pom => formatPomName(pom?.secondary_pom?.name))
+          .toPromiseSettledResult(),
+        communityOffenderManager: communityManager.map(formatCommunityManager).toPromiseSettledResult(),
+        resettlementWorker: bookingContacts.otherContacts
+          ?.filter(contact => contact.relationship === 'RW' && contact.activeFlag)
+          .sort((a, b) => sortByDateTime(b.createDateTime, a.createDateTime))
```
brightonsbox, line 299: null check before the filter but not on subsequent sort and map?

brightonsbox (approved):
LGTM

PR #1156, opened 2025-07-23
CDPS-1517 CDPS-1189 Implement row anchors on the personal page

scottrowley:


assets/scss/helpers.scss
```diff
@@ -209,3 +209,9 @@ div.hmpps-merged-key-summary-list-row {
 .hmpps-\!-float-right {
   float: right;
 }
+
```
scottrowley, line 212: Using standard anchors, the success flash message obscures some of the row that was updated, so I'm anchoring to to the centre of the screen (as suggested by content)

server/views/partials/personalPage/identityNumbers.njk
```diff
@@ -9,6 +9,7 @@
   {% if identityNumbers.length > 0 %}
     {% for identityNumber in identityNumbers %}
       <div class="identity-number-sub-row govuk-!-padding-top-2">
+        <span class="anchor-center" id="{{ options.redirectAnchor }}"></span>
```
scottrowley, line 12: Had to put this anchor here in the div holding the row "value", as doing it elsewhere messed up the table layout

brightonsbox (approved):
LGTM

PR #1169, opened 2025-07-30
CDPS-1530 Validate date of birth range

Ewan-Donovan:
Adds an extra layer of validation onto the dates used by alias controller. Checks the DOB is within 15 - 125 years ago.


brightonsbox (approved):
LGTM

PR #1206, opened 2025-09-03
CDPS-1540 Parallelise cypress tests

scottrowley:
We'll have to mess around with the expected checks in the repo config, but otherwise splitting into 3 has sped up the tests a lot.
Screenshot 2025-09-03 at 17 51 10


package.json
```diff
@@ -20,8 +20,11 @@
     "test:coverage": "jest --coverage",
     "test:ci": "jest --runInBand",
     "security_audit": "npx audit-ci --config audit-ci.json",
-    "int-test": "cypress run --config video=false",
```
scottrowley, line 23: e2e and video=false are both default values

PR #1217, opened 2025-09-10
CDPS-1618: Reduce sass deprecation warnings

ushkarev:
* Replace usage of deprecated functions with those from sass modules
* Reorder nested declarations (mainly from mixin includes) after plain ones to be compatible with future sass processing – this _does_ change the order of some built output declarations, but they don’t interact so the overall effect remains identical

What remains: use of `@import` all over the place. We cannot move until GOVUK _and_ MoJ frontends do so first. I tested a variety of `@use` combinations to try to reduce deprecated imports, to no avail. Had we _solely_ used GOVUK frontend, it’d actually work, but MoJ frontend can only work if the former is imported the old way. GDS don’t have a branch with any work on this yet so we just have to wait. At the very least, they will need to set a whole new namespace and `@use` convention that we can’t anticipate.


brightonsbox (approved):
LGTM!

PR #1252, opened 2025-09-25
Add personal officer entry action link to profile, and remove deprecated codes for key worker session action link

bill-moj:
MIA-900


ushkarev (approved):
LGTM, but somebody else probs should cast an eye over it…

PR #1268, opened 2025-10-03
CDPS-1394: Express 5 compatibility

ushkarev:
`asyncMiddleware` is not needed since express 5 can already deal with thrown exceptions inside request handlers. cf. [template project](https://github.com/ministryofjustice/hmpps-template-typescript/pull/560) removed it during the upgrade

### DIFF
--- a/server/errorHandler.test.ts
+++ b/server/errorHandler.test.ts
@@ -1,9 +1,9 @@
 import type { Express } from 'express'
 import request from 'supertest'
 import { DataAccess } from './data'
-import CommonApiRoutes from './routes/common/api'
 import { appWithAllRoutes } from './routes/testutils/appSetup'
 import { aliasServiceMock } from '../tests/mocks/aliasServiceMock'
+import { commonApiRoutesMock } from '../tests/mocks/comonApiRoutesMock'
 import AliasService from './services/aliasService'
 
 jest.mock('./services')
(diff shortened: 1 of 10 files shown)


server/routes/routerUtils.ts
```diff
@@ -1,20 +1,13 @@
 import { RequestHandler, Router } from 'express'
-import asyncMiddleware from '../middleware/asyncMiddleware'
 
 const getRequest =
```
ushkarev, line 3: these `getRequest`/`postRequest` helpers now don’t do anything other than hide the true interface of the router methods.

i really wanna get rid of them if there’s appetite…?

server/errorHandler.test.ts
```diff
@@ -14,7 +14,7 @@ beforeEach(() => {
   app = appWithAllRoutes({
     services: {
       dataAccess: {} as DataAccess,
-      commonApiRoutes: {} as CommonApiRoutes,
+      commonApiRoutes: commonApiRoutesMock(),
```
ushkarev, line 17: i had to introduce this mock because there were a few undefined request handlers in the `/api/*/image` routes that didn’t matter before as they didn’t reach express when wrapped with the async middleware. but now they are added to a real router and need to exist.

Ewan-Donovan (approved):

PR #1270, opened 2025-10-06
Removing production replica count config - overriden by autoscaling

brightonsbox:


helm_deploy/values-prod.yaml
brightonsbox, line 16: The presence of this config supersedes the replicaCount config so having both is a bit confusing

PR #1273, opened 2025-10-07
CDPS-1643 Add support for restricted alerts and remove hard-coded alert restrictions

scottrowley:


server/controllers/alertsController.ts
```diff
@@ -16,10 +16,11 @@ import Prisoner from '../data/interfaces/prisonerSearchApi/Prisoner'
 import getCommonRequestData from '../utils/getCommonRequestData'
 import { errorHasStatus } from '../utils/errorHelpers'
 
-// Some alert codes cannot be added or made inactive via the Prisoner Profile:
+// Some alert codes cannot be added or made inactive via the Prisoner Profile
+// The Alerts API now supports limiting alert administration to a list of named users so if a user list can be sourced
+// for these alerts then we can remove this check
 const excludedAlertCodes: string[] = [
   'DOCGM', // OCG Nominal - Do not share
-  'DRONE', // Drone Nominal - Do not share
```
scottrowley, line 22: We will look to move OCG Nominal to the user list system in the near future, but for now this check needs to remain in place.

PR #1275, opened 2025-10-07
Fix integration tests using wiremock to serve an image

ushkarev:
The image path `./../../assets/images/average-face.jpg` is supposed to be relative to the mock api typescript file, but wiremock cannot know that – so it’s always absent.

Cf. errors in all “Run wiremock” and “Run the node app” steps on CircleCI.
>  Error when working with FileSource:
> "Access to file ./../../assets/images/average-face.jpg is not permitted. An absolute path from the filesystem root might be specified"

> WARN HMPPS Prisoner Profile: Error calling Prison API, path: '/api/bookings/offenderNo/G6123VU/image/data?fullSizeImage=false', verb: 'GET' (responseStatus=500, data={}, message="Internal Server Error")

Using `base64Body ` in place of `bodyFileName` means that wiremock can simple reserve the content it was provided with.

### DIFF
--- a/integration_tests/mockApis/personIntegrationApi.ts
+++ b/integration_tests/mockApis/personIntegrationApi.ts
@@ -1,4 +1,10 @@
-import { stubGetWithBody, stubPatchWithResponse, stubPostWithResponse, stubPutWithResponse } from './utils'
+import {
+  loadFileAsBase64,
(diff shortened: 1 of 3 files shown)


brightonsbox (approved):
LGTM!

PR #1640, opened 2026-04-16
CDPS-1851 & CDPS-1852: Stop reporting uninteresting errors to Sentry

ushkarev:
CDPS-1851:
- Ignore timeouts and oauth access token issues; already logged in Application Insights
- Ignore 404s; also logged in Application Insights, but _feasibly_ could have been useful to keep to indicate when we’re not using a third-party api correctly

Also:
- CDPS-1852: Clean PII more thoroughly: replace prisoner numbers in URLs and address lookups
- Relay additional error context as often provided by DPS apis:
Relay additional error context as often provided by DPS apis


server/sanitisedError.ts
```diff
@@ -1,32 +0,0 @@
...
-  status?: number
-  headers?: unknown
-  data?: unknown
-  stack: string
-  message: string
-}
-
-export type UnsanitisedError = ResponseError
-
-export default function sanitise(error: UnsanitisedError, endpoint: string): SanitisedError {
```
ushkarev, line 15: now in https://github.com/ministryofjustice/hmpps-typescript-lib/blob/20a7109d36a00e7cd695b0a6f5901311fd6728fc/packages/rest-client/src/main/helpers/sanitiseError.ts#L7

Ewan-Donovan (approved):

PR #1648, opened 2026-04-21
CDPS-1914 & CDPS-1922: Improve Sentry.io error reporting

ushkarev:
- add breadcrumbs for all `superagent`-based requests
- report 404-type errors that are not thrown inside the application; ie. come from api calls
- anonymise reported error messages


nj963 (approved):
I have no issue with this

PR #1660, opened 2026-04-30
CDPS-1834: Update gov-uk frontend and moj frontend

Ewan-Donovan:


package.json
```diff
@@ -63,7 +63,7 @@
     "date-fns": "^4.1.0",
     "express": "^5.2.1",
     "express-session": "^1.19.0",
-    "govuk-frontend": "5.14.0",
+    "govuk-frontend": "6.1.0",
```
ushkarev, line 66: you can remove the 2 `TODO: remove once …` blocks in `.github/dependabot.yml` so that both frontend libs are kept updated in future

assets/scss/components/_status.scss
```diff
@@ -39,13 +39,13 @@
 
   &--prominent {
     background-color: govuk-tint(govuk-colour("blue"), 90%);
```
ushkarev: `govuk-tint` and `govuk-shade` are deprecated and we have a few around; they recommend swapping for appropriate variants of `govuk-colour`

assets/scss/components/_personal-distinguishing-marks.scss
```diff
@@ -11,7 +11,7 @@
   dl {
     margin: 0;
     &:not(:last-of-type) {
-      border-bottom: 1px solid $govuk-canvas-background-colour;
+      border-bottom: 1px solid govuk-colour("black", $variant: "tint-95");
```
ushkarev: as I understand their release notes, shouldn’t `$govuk-canvas-background-colour` now be `govuk-functional-colour(template-background)`?

ushkarev (commented):
are the template block overrides doing the right thing? cos apparently, `{% block header %}` is now `{% block govukHeader %}` and there are similar changes for main content and footer. i’m unsure about all the before/after start/end blocks

ushkarev (approved):

PR #1671, opened 2026-05-08
CDPS-1934: Update visibility checks to use new perms

Ewan-Donovan:
Tests remain unchanged since behaviour is unchanged, just now it goes through permissions lib.


brightonsbox (approved):
LGTM!

PR #1672, opened 2026-05-08
Cherry picks from the template

brightonsbox:


ushkarev (approved):
i suspect there might be some need to tweak bunyan and/or application insights stuff… but i don’t really know what it’s meant to do to check that it still does

PR #1674, opened 2026-05-11
CDPS-1913 Implemented tabbing & paged list macros

nj963:
- Implemented new shared macros: `pagedListWithCards.njk`, `filteredListPage.njk`, and `hmppsTabs.njk` for reusability.
- Refactored `alertsList.njk`, `caseNotesList.njk`, and `visitsDetailsPage/list.njk` to use `pagedListWithCards.njk`.
- Updated `prisonerCsraHistoryPage.njk`, `visitsDetails.njk`, `caseNotesPage.njk`, and `alertsPage.njk` to use `filteredListPage.njk`.


server/views/macros/filteredListPage.njk
```diff
@@ -0,0 +1,57 @@
...
+        {{ govukErrorSummary({
+            titleText: "There is a problem",
+            errorList: params.errors,
+            attributes: { 'data-qa-errors': true }
+        }) }}
+    {% endif %}
+
+    <div class="govuk-grid-row">
+        <div class="govuk-grid-column-full">
+            <h1{% if params.titleId %} id="{{ params.titleId }}"{% endif %} class="govuk-heading-l govuk-!-margin-top-5">{{ params.title | safe }}</h1>
```
ushkarev: `{{ params.title | safe }}` how come there’s the `safe` filter? I think that makes using this macro a little risky in the sense that we must know to pre-escape. if it’s _really_ necessary, then maybe the param should be `titleHtml` to make it more visible

server/views/macros/filteredListPage.njk
```diff
@@ -0,0 +1,57 @@
...
+                    {{ params.filterHtml | safe }}
+                </div>
+                <div class="govuk-grid-column-three-quarters hmpps-full-width-print">
+                    {{ caller() }}
+                </div>
+            </div>
+        {% else %}
+            <div class="govuk-grid-row">
+                <div class="govuk-grid-column-full" data-qa="{{ params.emptyStateDataQa }}">
+                    {{ params.emptyStateText | safe }}
```
ushkarev: can `emptyStateText` be renamed `emptyStateHtml` instead so that it’s clear that it won’t be escaped?

ushkarev (approved):

PR #1675, opened 2026-05-11
CDPS-1915: Handle location history when a location is not a cell

ushkarev:
the locations-inside-prison api returns 404 when attributes are requested for a non-cell location so “View details” on a location history row threw an error for occasional locations (eg. reception).


brightonsbox (approved):
LGTM

PR #1700, opened 2026-05-27
CDPS-1980 Update duplicate profile page content

nj963:


ushkarev (approved):
✅
fyi the `govuk-body` class doesn’t do anything on `p` elements because [global styles are on](https://github.com/ministryofjustice/hmpps-prisoner-profile/blob/main/assets/scss/index.scss#L1)

PR #1701, opened 2026-05-28
CDPS-1932: Announce flash messages

ushkarev:
Elements already on the page at load time that have “alert” role are *not* assertively read out ahead of other content. So this change dynamically rejigs the flash message element structure to have the text read out after a brief pause. Delays in the tens of milliseconds do not get noticed, so using a more generous 1s.

Last attempt was #1464

### DIFF
--- a/assets/scss/components/_flash-message.scss
+++ b/assets/scss/components/_flash-message.scss
@@ -1,10 +1,6 @@
 @keyframes flashMessageFadeOut {
   0% {
     opacity: 1;
-    visibility: visible;
-  }
-  80% {
-    opacity: 1;
   }
   100% {
     opacity: 0;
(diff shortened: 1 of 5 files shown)


assets/scss/components/_flash-message.scss
```diff
@@ -16,8 +12,7 @@
   z-index: 1;
   top: 75px;
   left: 50%;
-  opacity: 0;
```
ushkarev, line 19: transparency might interact with ‘visibility’ to screen readers

server/views/macros/hmppsFlashMessage.njk
```diff
@@ -0,0 +1,7 @@
+{% macro hmppsFlashMessage(params) -%}
+
+  <div class="hmpps-flash-message {% if params.anchored %}hmpps-flash-message--at-anchor{% endif %}" data-module="hmpps-flash-message">
+    <p role="alert">{{ params.text }}</p>
```
ushkarev, line 4: this `role` attribute does nothing, but it was already present and even the GDS error summary component includes it on load. no harm retaining it, i suppose.

Ewan-Donovan (approved):

PR #1720, opened 2026-06-08
CDPS-1984 Add yellow banner to prisoner profile for edit-the-profile functionality

nj963:
Adds a reusable yellow banner partial; implements one on the prisoner profile for edit-the-profile functionality.


assets/scss/components/_notification-banner-yellow.scss
```diff
@@ -0,0 +1,23 @@
+.notification-banner-yellow {
+  border: 1px solid #594d00;
+  padding: 8px 8px 6px;
+  margin-bottom: 15px;
+  background: #fff7bf;
+  @include govuk-font(16);
+
+  a,
+  a:visited {
```
ushkarev, line 9: might be a good idea to add `a:link` as well since it’s a styled pseudo-class

server/views/partials/notificationBannerYellow.njk
```diff
@@ -0,0 +1,7 @@
+<div class="notification-banner-yellow">
+  <p>
+    <span class="govuk-!-font-weight-bold">You can now {{ youCanNowAction }} in DPS.</span>
```
ushkarev, line 3: is it deliberate that this is bold without semantically being emphasised?

nj963: only in the sense that a lot of other places love using this instead of `<strong>` or what have you. not that there's *much* consistency, admittedly

assets/scss/components/_notification-banner-yellow.scss
nj963: solid spot, will update

ushkarev (approved):

PR #1756, opened 2026-06-25
CDPS-1960: Show x-ray body scan counts from DPS data only to DPS developers

ushkarev:
in non-production and only to DPS developers, the x-ray body scans section is repeated loading the counts from DPS instead of prison-api (NOMIS)

---

**NB: depends on https://github.com/ministryofjustice/hmpps-x-ray-body-scans-api/pull/30**

### DIFF
--- a/assets/scss/components/_xrays.scss
+++ b/assets/scss/components/_xrays.scss
@@ -29,4 +29,17 @@
       color: govuk-colour('white');
     }
   }
+
+  &.hmpps-xray--unsafe {
+    border: 0 dashed red;
+    border-top-width: 5px;
+    border-bottom-width: 5px;
+
+    &::before {
+      content: 'Do not use this information:';
+      display: block;
+      font-weight: bold;
+      color: red;
+    }
+  }
 }
--- a/server/data/interfaces/xRayBodyScansApi/index.ts
+++ b/server/data/interfaces/xRayBodyScansApi/index.ts
@@ -9,6 +9,8 @@ export interface ScanCountResponse {
   nomisCount: number
   dpsCount: number
   totalCount: number
+  fromScanDate: Date
+  toScanDate: Date
 }
 
 export interface XRayBodyScansApiClient {
--- a/server/data/interfaces/xRayBodyScansApi/index.ts
+++ b/server/data/interfaces/xRayBodyScansApi/index.ts
@@ -1,7 +1,7 @@
 export interface ScanCountRequest {
   prisonerNumber: string
-  fromStartDate?: Date | undefined
-  toStartDate?: Date | undefined
+  fromScanDate?: Date | undefined
+  toScanDate?: Date | undefined
 }
 
 export interface ScanCountResponse {
--- a/server/data/interfaces/xRayBodyScansApi/index.ts
+++ b/server/data/interfaces/xRayBodyScansApi/index.ts
@@ -1,0 +1,16 @@
+export interface ScanCountRequest {
+  prisonerNumber: string
+  fromStartDate?: Date | undefined
+  toStartDate?: Date | undefined
+}
+
+export interface ScanCountResponse {
+  prisonerNumber: string
+  nomisCount: number
+  dpsCount: number
+  totalCount: number
+}
+
+export interface XRayBodyScansApiClient {
+  countScans(request: ScanCountRequest): Promise<ScanCountResponse>
+}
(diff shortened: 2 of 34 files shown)


server/views/partials/personalPage/security.njk
```diff
@@ -1,11 +1,12 @@
...
+{% from "macros/summaryCardMacro.njk" import summaryCard %}
+
 {%- call summaryCard({title: "Security", id: "security"}) -%}
   <div class="govuk-grid-row" data-qa="security">
     <div class="govuk-grid-column-full">
       {% if security.xrays.total > 0 %}
         <div class="hmpps-xray govuk-!-margin-bottom-3" data-qa="security-xrays">
+          {# TODO: this 116 should not be hard-coded in a template file,
+                   rather api-driven or in config/business logic #}
           {% if security.xrays.total >= 116 %}
```
ushkarev, line 10: i think the do-not-scan business logic should be centralised, moved to the new api… thoughts??

assets/scss/components/_xrays.scss
```diff
@@ -29,4 +29,17 @@
...
+    border-top-width: 5px;
+    border-bottom-width: 5px;
+
+    &::before {
+      content: 'Do not use this information:';
+      display: block;
+      font-weight: bold;
+      color: red;
+    }
+  }
```
ushkarev, line 44: this might be a bit distracting even for an internal demo…? 🤣

server/data/interfaces/xRayBodyScansApi/index.ts
```diff
@@ -0,0 +1,21 @@
...
+  fromStartDate?: Date | undefined
+  toStartDate?: Date | undefined
+}
+
+export interface ScanCountResponse {
+  prisonerNumber: string
+  nomisCount: number
+  dpsCount: number
+  totalCount: number
+}
```
ushkarev, line 14: ~i think the api should echo back the filtered date range so the profile can simply display it rather than calculate the first-of-the-year and hope the default filter range matches.~ updated in https://github.com/ministryofjustice/hmpps-x-ray-body-scans-api/pull/30

similarly, i think it’d be good to add too-many-scans flag (see above). tho it’s not trivial since the specified date filter range many not be a year… i made a ticket for this CDPS-2024

server/data/interfaces/xRayBodyScansApi/index.ts
```diff
@@ -0,0 +1,21 @@
+export interface ScanCountRequest {
+  prisonerNumber: string
+  fromStartDate?: Date | undefined
+  toStartDate?: Date | undefined
```
ushkarev: not sure what the word “start” refers to here 🤔

ushkarev: changed in https://github.com/ministryofjustice/hmpps-x-ray-body-scans-api/pull/30

server/views/partials/personalPage/security.njk
```diff
@@ -1,11 +1,12 @@
...
-{% from "govuk/components/summary-list/macro.njk" import govukSummaryList %}
-{% from "govuk/components/warning-text/macro.njk" import govukWarningText %}
+{% from "macros/summaryCardMacro.njk" import summaryCard %}
+
 {%- call summaryCard({title: "Security", id: "security"}) -%}
   <div class="govuk-grid-row" data-qa="security">
     <div class="govuk-grid-column-full">
       {% if security.xrays.total > 0 %}
         <div class="hmpps-xray govuk-!-margin-bottom-3" data-qa="security-xrays">
+          {# TODO: this 116 should not be hard-coded in a template file,
```
MeliaBurnley, line 8: You're definitely not wrong :D 

We probably want to figure out where to put this in the future, we'll need it in a few places (XRBS service, WPIP, here, future reception service) so we could consider putting it in the API maybe?

server/views/partials/personalPage/security.njk
MeliaBurnley: I did not see your comment before I made mine, but yes I agree

MeliaBurnley (approved):

server/views/partials/personalPage/security.njk
ushkarev: i made a ticket CDPS-2024

server/views/partials/personalPage/security.njk
ushkarev: i created a ticket CDPS-2024

PR #1834, opened 2026-07-31
SDIT-4121: 🐛 Display religion history based on whether religion exists

petergphillips:


local-wiremock/mappings/prisoner-search.json
```diff
@@ -24,7 +24,7 @@
       "ethnicity": "White: Eng./Welsh/Scot./N.Irish/British",
       "youthOffender": true,
       "maritalStatus": "No",
-      "religion": "Celestial Church of God",
+      "religion": "Agnostic",
```
petergphillips, line 27: changing data so that religion from Prisoner Search is different from Prison API to show that it isn't being used

server/data/interfaces/prisonApi/Belief.ts
```diff
@@ -1,6 +1,4 @@
 export default interface Belief {
-  bookingId: number
```
petergphillips, line 2: removed as not used and wanting to change the API contract to match Core Person

server/services/personalPageService.ts
```diff
@@ -419,7 +417,7 @@ export default class PersonalPageService {
         ),
       },
       learnerNeurodivergence,
-      hasCurrentBelief: beliefs?.some(belief => belief.bookingId === bookingId),
+      hasCurrentBelief: inmateDetail.religion !== null,
```
petergphillips, line 420: Prison API sets the `religion` field now based off the current religion from history (rather than grabbing it from the profile as it used to do). We can therefore check for the presence of `religion` to determine if there is any history

server/services/personalPageService.ts
```diff
@@ -286,7 +285,6 @@ export default class PersonalPageService {
       !getOptions.editProfileEnabled ? prisonApiClient.getAddresses(prisonerNumber) : null,
       prisonApiClient.getOffenderContacts(prisonerNumber),
       prisonApiClient.getIdentifiers(prisonerNumber, getOptions.editProfileEnabled),
-      prisonApiClient.getBeliefHistory(prisonerNumber),
```
petergphillips, line 289: no need to get the belief history for the profile page now

MeliaBurnley (approved):
LGTM :)

PR #1835, opened 2026-07-31
SDIT-4084: 👽️ Remove bookingId as never used for religion history

petergphillips:


server/data/prisonApiClient.ts
```diff
@@ -545,11 +545,10 @@ export default class PrisonApiRestClient extends RestClient implements PrisonApi
     )
   }
 
-  async getBeliefHistory(prisonerNumber: string, bookingId?: number): Promise<Belief[]> {
+  async getBeliefHistory(prisonerNumber: string): Promise<Belief[]> {
     return this.get(
       {
         path: `/api/offenders/${prisonerNumber}/belief-history`,
-        query: bookingId ? `bookingId=${bookingId}` : undefined,
```
petergphillips, line 552: we're changing the Prison API endpoints to match the eventual Core Person endpoints and this was never passed in / used so removing

brightonsbox (approved):
LGTM

PR #1836, opened 2026-07-31
RR-2591 - Route and view changes to display Employability Skills from LWP

nathanrussell-moj-digital:
PR to update the Work & Skills route to retrieve Employability Skills from LWP, and to update the nunjucks template based on the new designs for Employability Skills:

If the person has any Employability Skills:
image

If the person has no Employability Skills:
image

(not shown in the images above, but the ticket specifies that "CIAG" should be spelt out as "The careers information, advice and guidance (CIAG)" )

Whether to retrieve and render Employability Skills from Curious or LWP is based on the feature toggle. The feature toggle is still disabled in all environments.
The next PR will update the cypress tests, and enable the toggle in `dev`


brightonsbox (approved):
LGTM

PR #1840, opened 2026-08-03
CDPS-2057: Update content when no confirmed release dates are recorded

ushkarev:


Ewan-Donovan (approved):
Yep this looks more appropriate

PR #1841, opened 2026-08-03
CDPS-1974 Improve breadcrumbs in the profile

nj963:
Lots of small changes made to multiple files here; essentially just a refactor. Most helpful change is in `index.njk`

### DIFF
--- a/server/views/pages/prisonerSchedule.njk
+++ b/server/views/pages/prisonerSchedule.njk
@@ -3,19 +3,7 @@
 
 {% extends "./index.njk" %}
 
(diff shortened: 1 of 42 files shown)


server/views/pages/prisonerSchedule.njk
```diff
@@ -3,20 +3,8 @@
 
 {% extends "./index.njk" %}
 
-{% set title = 'Schedule' %}
```
ushkarev, line 6: is the title set in a different manner as well?

ushkarev (approved):

PR #1876, opened 2026-08-19
MAPA-281 Read the cell move reason from the cell movements API

Mjwillis:
Switches the location history page's "what happened" off whereabouts-api. The page previously made two hops against dying services — whereabouts for a case note id, then offender-case-notes for its text. It now makes **one call** to `hmpps-change-someones-cell-api`:

`GET /cell-movements/{bookingId}/bed-assignment/{bedAssignmentSequence}` → `commentText` is the explanation.

The new API serves movements it recorded itself, movements already migrated from whereabouts, and — transitionally — movements still only in whereabouts, which it migrates on first read. So this page keeps answering for all history with no cutover gap.

## Changes

- New `cellMovementsApiClient` (`CELL_MOVEMENTS_API_URL`, health ping included) following the house client pattern; `getCellMovementReason(bookingId, seq, ignore404)` with the same 404-tolerant behaviour as before — nothing recorded still renders "Not entered".
- `prisonerLocationHistoryService.fetchWhatHappened` collapses to the single call; the service drops its whereabouts and case-notes dependencies.
- `getCellMoveReason` removed from the whereabouts client and its interface — **appointments are now this app's only whereabouts use** (plus a dead `getUnacceptableAbsences` method nothing calls, left for a separate tidy-up).
- "Reason for move" behaviour is unchanged: still sourced from prison-api's bed-assignment history + `CHG_HOUS_RSN` reference data, still only shown when an explanation exists.
- Cypress: whereabouts cell-move-reason stub replaced by a cell movements API stub; the location history spec passes with no assertion changes. Health ping added to `setupHealthPings`.

## FYI

The app needed **system client needs `ROLE_CELL_MOVEMENTS__RO`** in HMPPS Auth (all envs).  I've added them.


Mjwillis:
Added a second commit covering the ticket's other acceptance criterion: "Reason for move" is no longer gated on the explanation existing — the reason code comes from prison-api's bed assignment history and now renders even when no case note/explanation was ever recorded (previously both silently showed "Not entered"). New controller test proves the decoupling.

nj963 (approved):
LGTM 👍

server/data/cellMovementsApiClient.ts
```diff
@@ -0,0 +1,36 @@
...
+   * the whereabouts-api + case-notes two-hop this page made before; 404 means nothing was recorded
+   * through DPS for this bed assignment, which is the common case and matches what whereabouts
+   * returned.
+   *
+   * Requires the system client to hold ROLE_CELL_MOVEMENTS__RO.
+   */
+  async getCellMovementReason(
+    bookingId: number,
+    bedAssignmentSequence: number,
+    ignore404: boolean,
```
brightonsbox, line 22: It looks like this is always called with `ignore404` set to `true`?

